PR #3, opened 2018-10-29
Add diamond node

### LINKED ISSUE #2: Add diamond node with proper anchoring of edges
Several diagram notations, such as flow charts, use nodes shaped in the form of a diamond / rhombus with an independent width and height. This new node should provide proper anchoring support of connected edges taking the shape of a diamond and its height and width into account.

This issue has been originally raised and discussed in https://github.com/theia-ide/sprotty/issues/235.

planger:
This change introduces a diamond node to the node library with proper anchoring of edges. Closes #2.

### DIFF
--- a/src/utils/geometry.ts
+++ b/src/utils/geometry.ts
@@ -380,12 +380,9 @@ export class Diamond {
  * A line represented in its standard form `a*x + b*y = c`.
  */
 export interface Line {
-
     readonly a: number
     readonly b: number
     readonly c: number
-
-    intersection(otherLine: Line): Point
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/utils/geometry.ts
```diff
@@ -260,3 +298,119 @@ export function toRadians(a: number): number {
...
+
+    get b(): number {
+        return this.p2.x - this.p1.x;
+    }
+
+    get c(): number {
+        return this.p2.x * this.p1.y - this.p1.x * this.p2.y;
+    }
+
+    intersection(other: Line): Point {
```
spoenemann: This intersection function is not specific to `PointToPointLine` because it does not use `p1` or `p2`, but just `a`, `b` and `c`. It could be extracted to a global function (and removed from the interface). That would also allow any JSON object with three properties `a`, `b` and `c` to be used as `Line`.

planger: Right, thanks! I factored out the function in 9a302a7.

PR #31, opened 2018-12-20
#30: Support 'keyup' event in KeyTool

### LINKED ISSUE #30: Support 'keyup' event in KeyTool
The KeyTool currently only supports listening to the 'keydown' event but we would also need to listen to the  'keyup' event.

martin-fleck-at:
Fixes #30 by supporting 'keyup' event

Signed-off-by: Martin Fleck <mfleck@eclipsesource.com>


spoenemann:
Thank you!

PR #32, opened 2019-01-07
Add IActionDispatcherProvider type

planger:
The IActionDispatcherProvider type has been missing, however, it is available in the DI container. Thus, to make it easier for clients to obtain the action dispatcher via a provider, this change adds and exports the corresponding type.


JanKoehnlein:
Thanks for the patch.

PR #35, opened 2019-01-14
[edges] consider edges when calculating diagram bounds

### LINKED ISSUE #9: Edges are not considered by zoom-to-fit
From theia-ide/sprotty#219:

> When applying zoom-to-fit, edges may not be visible because they are not bounds-aware and thus not considered. This problem might apply to other features, too (e.g. SVG export).

JanKoehnlein:
Fixes #9

### DIFF
--- a/src/graph/sgraph.ts
+++ b/src/graph/sgraph.ts
@@ -16,7 +16,7 @@
 
 import { SChildElement, SModelElement, SModelElementSchema, SModelIndex, SModelRootSchema } from '../base/model/smodel';
 import { Alignable, alignFeature, boundsFeature, layoutableChildFeature, layoutContainerFeature, ModelLayoutOptions,
-    SShapeElement, SShapeElementSchema } from '../features/bounds/model';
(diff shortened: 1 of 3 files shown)


src/graph/sgraph.ts
```diff
@@ -125,7 +125,7 @@ export interface SEdgeSchema extends SModelElementSchema {
  * each of which can be either a node or a port. The source and target elements are referenced via their
  * ids and can be resolved with the index stored in the root element.
  */
-export class SEdge extends SChildElement implements Fadeable, Selectable, Routable, Hoverable {
+export class SEdge extends SChildElement implements Fadeable, Selectable, Routable, Hoverable, BoundsAware {
```
spoenemann: Can this lead to other side-effects? The BoundsAware interface is used in several places.

JanKoehnlein:
I was asking myself the same question, but so far I didn't encounter any issues during testing. Usually there are further guards, e.g. `isLayoutable()`, before more in depth handling.

Ha! Just found out that I broke the resize on update feature. Fixed it.

PR #37, opened 2019-01-15
[examples] cleaned up examples

### LINKED ISSUE #36: [examples] Fix the examples
They still contain the java stuff, and they are not usable (SVG) without a server.
Mindmap doesn't work at all.

JanKoehnlein:
- made the examples run again by adding a http server
- load CSS from node_modules in TS code
- removed Java examples
- run examples from gitpod
- fixed mindmap popup
- fixed mindmap popup button rendering

Fixes #36


JanKoehnlein:
I load the sprotty css in TypeScript now, as the http-server does not allow to access the parent directory.

PR #40, opened 2019-01-22
Fix zoom level computation in FitToScreenCommand (#39)

### LINKED ISSUE #39: FitToScreenAction with padding leads to wrong zoom level
If a `FitToScreenAction` with a padding is dispatched, the zoom level is wrong regarding the height. It looks like parenthesis are missing:
https://github.com/eclipse/sprotty/blob/579bee34577f2243a36fc178c6bef9debc9116ea/src/features/viewport/center-fit.ts#L198-L200

I'll open a PR in a sec.

planger:
Fix #39


spoenemann:
Thank you!

Issue #45, opened 2019-01-31
[di] consider creating commands and views using DI


JanKoehnlein:
Currently, we create views and commands directly using the constructor. That causes problems when you want to use a specific service, that is usually provided (and as such configurable) using DI. Our current workarounds are
1) Add the service to the `RenderingContext`/`CommandExecutionContext`. While this is OK for services from Sprotty's `base` modules (e.g. ViewRegistry), but causes a dependency cycle when a downstream module wants to contribute something (e.g. `graph` or some `feature`).
2) Add the service to the `SModel`, as we did for `SEdge#route()` or the anchor calculation. This turns the internal model into an object-oriented API, while the rest of Sprotty rather follows a service-oriented/functional pattern. We're loosing all the nice declarative properties of the SModel.

In the particular case I am currently trying to solve, I want to provide different edge routing strategies based on a property of the edge. I need these strategies in the view as well as in the command, but there is currently no way to inject a `EdgeRouterRegistry` into these. 

If we create commands and views using DI, we will solve all of the problems generically and have a much clearer architecture. Chances are high that we can hide the heavy-lifting with InversifyJS behind existing Sprotty APIs. 

The API changes would include
* the registration of commands (maybe)
* make all commands and Views `@injectable`
* turning anchor calculation into a service

PR #46, opened 2019-02-01
#45 make views and commands injectable

JanKoehnlein:
Also makes views singletons.

### DIFF
--- a/src/base/commands/command-registration.ts
+++ b/src/base/commands/command-registration.ts
@@ -1,4 +1,3 @@
-
 /********************************************************************************
  * Copyright (c) 2019 TypeFox and others.
  *
@@ -54,6 +53,9 @@ export interface ICommandConstructor {
     new (a: Action, ...args: any[]): ICommand
 }
 
+/**
+ * Use this method in your DI configuration to register a new command to the diagram.
+ */
 export function configureCommand(context: { bind: interfaces.Bind, isBound: interfaces.IsBound }, constr: ICommandConstructor) {
     if (isInjectable(constr)) {
         if (!context.isBound(constr))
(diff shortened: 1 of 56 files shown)


JanKoehnlein:
Breaking changes in this PR
- views are now singletons
- views have to be @injectable
- views can no longer be registered to the viewRegistry directly
- views have to be configured with the new `configureView({bind, isBound}, 'mytype', MyView)` or with `configureModelElement`
- configureModelElement has a new parameter property `isBound`
- commands have to be @injectable
- actions passed to command constructors have to be annotated with `@inject(TYPES.Action)`
- commands can no longer be registered to the `ActionHandlerRegistry` directly, e.g. in model sources
- commands have to be registered with `configureCommand({bind, isBound}, MyCommand)` in the DI container now
- the SetModelCommand is now registered in the base DI config
- the UpdateModelCommand is now registered using the `updateModule`

src/base/commands/command-registration.ts
```diff
@@ -0,0 +1,75 @@
...
+    kind: string
+    factory: (a: Action) => ICommand
+}
+
+export interface ICommandConstructor {
+    KIND: string
+    new (a: Action, ...args: any[]): ICommand
+}
+
+export function configureCommand(context: { bind: interfaces.Bind, isBound: interfaces.IsBound }, constr: ICommandConstructor) {
```
spoenemann: Please add a little comment for documentation.

planger:
Very nice improvement! Thanks! With commands, we ran into this issue already from time to time. I'm sure this will allow us to improve the code in GLSP.

> @planger do you have any concerns regarding GLSP (apart from the effort to adapt to these API changes)?

Thanks for asking! No, we appreciate this change! We will adapt GLSP and report back if there are any issues that appear in the process.

PR #47, opened 2019-02-01
[tests] switched to JS based testing (no more tsnode)

JanKoehnlein:
I had serious problems running and debugging the tests from TS, so I switched to JS based testing and got rid of ts-node.


JanKoehnlein:
I think this is outdated.

Issue #58, opened 2019-02-12
[api] remove strange APIs


JanKoehnlein:
Remove 
* `SModelStorage`: SModels are often to big to be put on browser local storage.
* `DiagramState` / `ExpansionState`: dirty UI state can just be stored in the SModel (internal model) now
* `IStateAwareModelProvider`: doesn't make sense anymore

PR #60, opened 2019-02-12
API fixes

JanKoehnlein:
- introduced commit action #6
- got rid of strange APIs #58


src/model-source/commit-model.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Commit the current SModel back to the model source.
+ *
+ * The SModel (AKA internal model) contains a lot of dirty/transitional state, such
+ * as intermediate move postions or handles. When a user interaction that spans multiple
+ * commands finishes, it fires a CommitModelAction to write the final changes back to
+ * the model source.
+ */
+export class CommitModelAction {
```
spoenemann: `implements Action`

src/model-source/commit-model.ts
```diff
@@ -0,0 +1,65 @@
...
+    kind = CommitModelCommand.KIND;
```
spoenemann: `readonly`

src/model-source/commit-model.ts
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@injectable()
+export class CommitModelCommand extends Command {
+    static KIND = 'commitModel';
+
+    @inject(TYPES.IModelFactory) modelFactory: IModelFactory;
```
spoenemann: The factory is currently part of the CommandExecutionContext. Should we remove it from there so every command can inject it if required? Same for the ILogger.

src/model-source/model-source.ts
```diff
@@ -58,4 +59,15 @@ export abstract class ModelSource implements IActionHandler {
     }
 
     abstract handle(action: Action): ICommand | Action | void;
+
+    /**
+     * Commit changes from the internal SModel back to the currentModel.
```
spoenemann: Please add something like: "This method is meant to be called only by `CommitModelCommand` and other commands that need to feed the current internal model back to the model source."

src/model-source/model-source.ts
```diff
@@ -58,4 +59,15 @@ export abstract class ModelSource implements IActionHandler {
...
+     *
+     * Does not have any side effects such as triggering layout or bounds computation,
+     * as the internal model is already current. See <code>CommitModelAction</code>
```
spoenemann: Minor: we usually use Markdown for formatting comments.

Issue #59, opened 2019-02-12
[api] get rid of constructor injection


JanKoehnlein:
Constructor injection induces much more API breaks than necessary. We should prefer property injection whenever applicable.

PR #63, opened 2019-02-13
Add support for a command palette (#34)

### LINKED ISSUE #34: Support a command palette to quickly invoke actions
With https://github.com/eclipsesource/graphical-lsp/pull/136 we proposed a command palette for GLSP. In GLSP we will fill this command palette with editing actions that come from the server (similar to code completion in LSP). However the base is independent of that as it obtains all available actions from a dedicated DI-configurable action provider.

Is that something that you would welcome in Sprotty as well? If yes, we could extract the GLSP-independent part and move this to sprotty.

Here a short screen cast how the initial implementation looks like:
command-palette-demo

planger:
Introduces an extensible command palette as a keyboard-oriented way of invoking actions similar to a Code Completion / IntelliSense / Quick-Fix functionality for code. The available commands are determined by registered action providers that provide a set of `LabeledAction`s for a context. A `LabeledAction` is essentially a label to be shown in the list and an array of Sprotty `Action`s to be dispatched on invocation.

The command palette is shown on <kbd>Ctrl</kbd> + <kbd>Space</kbd>. If there is no selection or multi-selection, it shows up in the left-upper corner of the base div. If there is a single selection, it is shown on top of the selected element.

The command palette is built on top of a UI extension mechanism introduced with this PR as well to register UI extensions. UI extensions allow to show additional UI elements (HTML elements) on top of Sprotty diagrams. Such UI extensions can be registered with an ID and then enabled or disabled by dispatching respective actions for showing and hiding extensions.
A UI extension is rendered in a separate div element but can be positioned on top of Sprotty diagrams based on a current context (e.g. selected elements, etc.).

In near future, we can implement a graphical palette on top of the same UI extension basis easily (as is currently done in GLSP).

Finally, for demonstration purposes, we integrated the introduced a command palette with a simple action provider in Sprotty's class diagram example.

Fixes #34.

### DIFF
--- a/src/base/ui-extensions/ui-extension-registry.ts
+++ b/src/base/ui-extensions/ui-extension-registry.ts
@@ -1,0 +1,112 @@
+/********************************************************************************
+ * Copyright (c) 2019 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { inject, injectable, multiInject, optional } from "inversify";
+import { InstanceRegistry } from "../../utils/registry";
+import { Action } from "../actions/action";
+import { ActionHandlerRegistry, IActionHandler, IActionHandlerInitializer } from "../actions/action-handler";
+import { CommandExecutionContext, CommandResult, ICommand, SystemCommand } from "../commands/command";
+import { TYPES } from "../types";
+import { IUIExtension } from "./ui-extension";
+
+/**
+ * Action requesting to show the UI extension with the specified `id`.
+ */
+export class ShowUIExtensionAction implements Action {
+    static KIND = "showUIExtension";
+    readonly kind = ShowUIExtensionAction.KIND;
+    constructor(public readonly extensionId: string) { }
+}
+
+/**
+ * Action requesting to hide the UI extension with the specified `id`.
+ */
+export class HideUIExtensionAction implements Action {
+    static KIND = "hideUIExtension";
+    readonly kind = HideUIExtensionAction.KIND;
+    constructor(public readonly extensionId: string) { }
+}
+
+/**
+ * The registry maintaining UI extensions registered via `TYPES.IUIExtension`.
+ */
+@injectable()
+export class UIExtensionRegistry extends InstanceRegistry<IUIExtension>  {
+    constructor(@multiInject(TYPES.IUIExtension) @optional() extensions: (IUIExtension)[] = []) {
+        super();
+        extensions.forEach((extension) => this.register(extension.id, extension));
+    }
+}
+
+/**
+ * Initalizer and handler for actions related to UI extensions.
+ */
+@injectable()
+export class UIExtensionActionHandlerInitializer implements IActionHandlerInitializer, IActionHandler {
+
+    @inject(TYPES.UIExtensionRegistry) protected readonly registry: UIExtensionRegistry;
+
+    initialize(registry: ActionHandlerRegistry): void {
+        registry.register(ShowUIExtensionAction.KIND, this);
+        registry.register(HideUIExtensionAction.KIND, this);
+    }
+
+    handle(action: Action): void | ICommand | Action {
+        if (action instanceof ShowUIExtensionAction) {
+            return new UIExtensionActionCommand((context) => {
+                this.withExtension(action.extensionId, (extension) => {
+                    extension.show(context);
+                });
+            });
+        } else if (action instanceof HideUIExtensionAction) {
+            return new UIExtensionActionCommand((context) => {
+                this.withExtension(action.extensionId, (extension) => {
+                    extension.hide();
+                });
+            });
+        }
+    }
+
+    protected withExtension(extensionId: string, func: (extension: IUIExtension) => void) {
+        const extension = this.registry.get(extensionId);
+        if (extension) {
+            func(extension);
+        }
+    }
+}
+
+/**
+ * A system command that doesn't change the model but just performs a specified `effect`.
+ */
+@injectable()
+export class UIExtensionActionCommand extends SystemCommand {
+
+    constructor(readonly effect: (context: CommandExecutionContext) => void) {
+        super();
+    }
+
+    execute(context: CommandExecutionContext): CommandResult {
+        this.effect(context);
+        return context.root;
+    }
+
+    undo(context: CommandExecutionContext): CommandResult {
+        return context.root;
+    }
+
+    redo(context: CommandExecutionContext): CommandResult {
+        return context.root;
+    }
+}
--- a/src/base/ui-extensions/ui-extension-registry.ts
+++ b/src/base/ui-extensions/ui-extension-registry.ts
@@ -16,28 +16,9 @@
 import { inject, injectable, multiInject, optional } from "inversify";
 import { InstanceRegistry } from "../../utils/registry";
 import { Action } from "../actions/action";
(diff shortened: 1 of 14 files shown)


planger:
Two things, I'd like to highlight for discussion:
1. We add a dependency to https://github.com/kraaden/autocomplete (MIT License). Thus, we might need to raise a CQ with the IP team. If you prefer to use a different auto-completion component, please let me know.
2. Currently we pass the `CommandExecutionContext` to the UI extensions. This allows them to access the index and the root element. It might be worth considering introducing a dedicated interface for that purpose or to only pass the root element? Please let me know what you think.

Thank you!

src/base/ui-extensions/ui-extension-registry.ts
```diff
@@ -0,0 +1,112 @@
...
+        } else if (action instanceof HideUIExtensionAction) {
+            return new UIExtensionActionCommand((context) => {
+                this.withExtension(action.extensionId, (extension) => {
+                    extension.hide();
+                });
+            });
+        }
+    }
+
+    protected withExtension(extensionId: string, func: (extension: IUIExtension) => void) {
```
spoenemann: With the changes made my @JanKoehnlein it should be possible to inject the UIExtensionRegistry directly in the command, so the code could be simplified. It seems strange that the command accepts an arbitrary function.

src/features/nameable/model.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function isNameable(element: SModelElement): element is SModelElement & Nameable {
+    return element.hasFeature(nameFeature);
+}
+
+export function name(element: SModelElement): string {
+    if (isNameable(element)) {
+        return element.name;
+    } else {
+        return 'unnamed';
```
spoenemann: Wouldn't it be more appropriate to return `undefined` in this case?

examples/classdiagram/src/model-source.ts
```diff
@@ -145,6 +145,7 @@ export class ClassDiagramModelSource extends LocalModelSource {
         const node0 = {
             id: 'node0',
             type: 'node:class',
+            name: 'Foo',
```
spoenemann: This leads to redundant information in the model: the name is present in the `name` attribute and in the text of the `label:heading` type label. This might be a problem when one wants to rename an element. It could be solved in this case by replacing `name` with a getter property that retrieves the value of the heading label.

src/features/command-palette/command-palette.ts
```diff
@@ -0,0 +1,180 @@
...
+                    y += (firstElement.bounds.y - viewport.scroll.y) * viewport.zoom;
+                } else {
+                    x += firstElement.bounds.x;
+                    y += firstElement.bounds.y;
+                }
+            }
+        }
+        containerElement.style.left = `${x}px`;
+        containerElement.style.top = `${y}px`;
+        containerElement.style.width = `${this.defaultWidth}px`;
```
spoenemann: The input box should appear at the side of the selected element, not on top of it.

src/features/command-palette/di.config.ts
```diff
@@ -0,0 +1,35 @@
...
+import { TYPES } from "../../base/types";
+import { CommandPaletteActionProviderRegistry, ICommandPaletteActionProvider } from "./action-providers";
+import { CommandPalette, CommandPaletteKeyListener } from "./command-palette";
+
+const commandPaletteModule = new ContainerModule((bind) => {
+    bind(CommandPalette).toSelf().inSingletonScope();
+    bind(TYPES.IUIExtension).toService(CommandPalette);
+    bind(TYPES.KeyListener).to(CommandPaletteKeyListener);
+    bind(CommandPaletteActionProviderRegistry).toSelf().inSingletonScope();
+    bind(TYPES.ICommandPaletteActionProviderRegistry).toProvider<ICommandPaletteActionProvider>((context) => {
```
spoenemann: Why isn't this just a `toService` binding?

src/features/command-palette/command-palette.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    protected paletteContext?: CommandExecutionContext;
+    protected contextActions?: LabeledAction[];
+
+    constructor(
+        @inject(TYPES.ViewerOptions) protected options: ViewerOptions,
+        @inject(TYPES.IActionDispatcherProvider) protected actionDispatcherProvider: IActionDispatcherProvider,
+        @inject(TYPES.ICommandPaletteActionProviderRegistry) protected actionProvider: ICommandPaletteActionProviderRegistry,
+        @inject(TYPES.ILogger) protected logger: ILogger) {
+        super(options, logger);
```
spoenemann: Please consider using property injection instead of constructor injection (see #59)

src/features/command-palette/action-providers.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export interface ICommandPaletteActionProvider {
+    getActions(context: CommandExecutionContext): Promise<LabeledAction[]>;
+}
+
+export type ICommandPaletteActionProviderRegistry = () => Promise<ICommandPaletteActionProvider>;
+
+@injectable()
+export class CommandPaletteActionProviderRegistry implements ICommandPaletteActionProvider {
+    public actionProvider: ICommandPaletteActionProvider[] = [];
```
spoenemann: Rename to `actionProviders`.

src/base/ui-extensions/ui-extension.ts
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    /**
+     * Updates the `containerElement` under the given `context` before it becomes visible.
+     *
+     * Subclasses may override this method to, for instance, modfying the position of the
+     * `containerElement`, add or remove elements, etc. depending on the specified `context`.
+     */
+    protected updateBeforeBecomingVisible(containerElement: HTMLElement, context: CommandExecutionContext): void {
```
spoenemann: This method name could be shortened, e.g. `onBeforeShow`.

src/base/ui-extensions/ui-extension.ts
```diff
@@ -0,0 +1,120 @@
...
+        // default: do nothing
+    }
+
+    /**
+     * Initializes the contents of this UI extension.
+     *
+     * Subclasses must implement this method to initialize the UI elements of this UI extension inside the specified `containerElement`.
+     */
+    protected abstract initializeUIExtension(containerElement: HTMLElement): void;
```
spoenemann: Maybe rename to `initializeContents`?

src/base/ui-extensions/ui-extension.ts
```diff
@@ -0,0 +1,120 @@
...
+    abstract readonly id: string;
+    abstract readonly containerClass: string;
+    protected containerElement: HTMLElement;
+    protected activeElement: Element | null;
+
+    constructor(
+        @inject(TYPES.ViewerOptions) protected options: ViewerOptions,
+        @inject(TYPES.ILogger) protected logger: ILogger) { }
+
+    show(context: CommandExecutionContext): void {
```
spoenemann: It should be sufficient to pass the root element (maybe marked with `Readonly<>` to indicate that it shouldn't be modified here). The CommandExecutionContext is related to command execution and viewer updates, and should not be accessed in this context.

planger:
Thank you very much for your review, @spoenemann! The feedback is great and we will incorporate your comments in a follow-up commit on this PR soon. Thanks again!

tortmayr:
Fixed the compilation errors. However, the ECA check seems to fail although I signed my commits.

planger:
Hm, I suspected that the ECA check fails because the branch contained merge commits. However, now I rebased everything on top of master and the ECA check still shows up, even though all additional commits include the sign-off. Not sure how to make this check happy, I'm sorry! :/

src/base/ui-extensions/ui-extension.ts
```diff
@@ -0,0 +1,118 @@
...
+                this.containerElement.style.visibility = 'hidden';
+                this.containerElement.style.opacity = '0';
+            }
+        }
+    }
+
+    /**
+     * Updates the `containerElement` under the given `context` before it becomes visible.
+     *
+     * Subclasses may override this method to, for instance, modfying the position of the
```
spoenemann: Minor: `modfying` => `modifying`

planger:
@spoenemann Thanks a lot for the review! I agree, we should add a keylistener to close any UI extension on <kbd>Esc</kbd>. We'll open an issue and handle this in a follow-up PR.

JanKoehnlein:
Sorry for the late review. LGTM. Please go on with the CQ

spoenemann:
https://dev.eclipse.org/ipzilla/show_bug.cgi?id=19395

planger:
Thanks a lot for merging this branch! I opened up #76 re closing the command palette on <kbd>Esc</kbd>.

PR #65, opened 2019-02-15
#49 make sprotty importable without 'lib'

### LINKED ISSUE #49: Declare 'main' and 'types' in package.json
We should have something like this

```
  "main": "lib/index",
  "types": "lib/index"
```

so consumers of our library can import from `sprotty` instead of `sprotty/lib`.

See https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

JanKoehnlein:
Fixes #49

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,5 +103,7 @@
     "lib",
     "css",
     "src"
-  ]
+  ],
+  "main": "lib/index",
+  "types": "lib/index"
 }


JanKoehnlein:
Sorry, I am merging this to reach a stable state with sprotty and sprotty-theia. Your feedback is appreciated nevertheless. Let's apply potential changes in separate PRs.

PR #77, opened 2019-03-23
Hide command palette when Escape key pressed

### LINKED ISSUE #76: Directly close command palette when user presses Escape key
The command palette introduced with #34 only disappears on blur (e.g. if the user clicks somewhere else). Especially since this is a keyboard-oriented way of invoking commands, we should also allow to hide the palette on <kbd>Esc</kbd>.

planger:
Fixes #76


planger:
Thanks for the review!

PR #82, opened 2019-04-25
Add more reliable absolute coordinates computation

### LINKED ISSUE #81: Find a reliable way to determining client-absolute coordinates of SModelElements
I realized that the command palette is placed incorrectly in different contexts. This can be reproduced, for instance, by removing the bootstrap css import from the `class-diagram.html` example.

This issue caused by the offsets of the parents of sprotty's base div so that the absolute positioning either takes the browser's page as their coordinate system's origin or sprotty's base div.

We should therefore find a computation of absolute positions of e.g. UI extensions that works for any HTML context the Sprotty base div might be placed in.

planger:
This computation is available in the function `getAbsoluteClientBounds` which is now also used for computing the absolute position of the command palette.

Fixes #81

### DIFF
--- a/src/features/bounds/model.ts
+++ b/src/features/bounds/model.ts
@@ -14,10 +14,12 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { Bounds, EMPTY_BOUNDS, EMPTY_DIMENSION, Dimension, isBounds, ORIGIN_POINT, Point, includes } from "../../utils/geometry";
-import { SModelElement, SModelElementSchema, SParentElement, SChildElement, SModelRoot } from "../../base/model/smodel";
+import { SChildElement, SModelElement, SModelElementSchema, SModelRoot, SParentElement } from "../../base/model/smodel";
 import { SModelExtension } from "../../base/model/smodel-extension";
 import { findParentByFeature } from '../../base/model/smodel-utils';
+import { DOMHelper } from "../../base/views/dom-helper";
+import { ViewerOptions } from "../../base/views/viewer-options";
+import { Bounds, Dimension, EMPTY_BOUNDS, EMPTY_DIMENSION, includes, isBounds, ORIGIN_POINT, Point } from "../../utils/geometry";
 import { Locateable } from '../move/model';
 
 export const boundsFeature = Symbol('boundsFeature');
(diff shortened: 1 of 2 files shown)


src/features/command-palette/command-palette.ts
```diff
@@ -82,16 +84,15 @@ export class CommandPalette extends AbstractUIExtension {
...
-            if (isBoundsAware(firstElement)) {
-                const viewport = findParentByFeature(firstElement, isViewport);
-                if (viewport) {
-                    x += (firstElement.bounds.x + firstElement.bounds.width - viewport.scroll.x) * viewport.zoom;
-                    y += (firstElement.bounds.y - viewport.scroll.y) * viewport.zoom;
-                } else {
-                    x += firstElement.bounds.x + firstElement.bounds.width;
-                    y += firstElement.bounds.y;
-                }
+            if (firstElement) {
```
spoenemann: Why should `selectedElements[0]` ever be undefined?

src/features/bounds/model.ts
```diff
@@ -96,6 +98,50 @@ export function getAbsoluteBounds(element: SModelElement): Bounds {
...
+    let height = 0;
+
+    const svgElementId = domHelper.createUniqueDOMElementId(element);
+    const svgElement = document.getElementById(svgElementId);
+    if (svgElement) {
+        const rect = svgElement.getBoundingClientRect();
+        const body = document.body;
+        const docElement = document.documentElement;
+        const scrollTop = window.pageYOffset || docElement.scrollTop || body.scrollTop;
+        const scrollLeft = window.pageXOffset || docElement.scrollLeft || body.scrollLeft;
```
spoenemann: Why not use `window.scrollX` and `window.scrollY` as [documented on MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)? Maybe then you wouldn't need to take into account the `clientTop` / `clientLeft` values.

src/features/bounds/model.ts
```diff
@@ -96,6 +98,50 @@ export function getAbsoluteBounds(element: SModelElement): Bounds {
...
+        const clientTop = docElement.clientTop || body.clientTop || 0;
+        const clientLeft = docElement.clientLeft || body.clientLeft || 0;
+
+        x = rect.left + scrollLeft - clientLeft;
+        y = rect.top + scrollTop - clientTop;
+        width = rect.right - rect.left;
+        height = rect.bottom - rect.top;
```
spoenemann: `rect.width` / `rect.height`

src/features/command-palette/command-palette.ts
planger: Oh, right, good catch. Thanks!

src/features/bounds/model.ts
planger: You're right, this seems to work too. Thanks!

planger:
@spoenemann Thanks for your review, Miro! I force-pushed an updated commit addressing your feedback!

Issue #83, opened 2019-04-26
Add user interface for label editing


planger:
For use cases of Sprotty that support editing, it would be nice to also have a UI allowing the user to conveniently edit labels directly in the diagram. I've created an initial implementation that adds a new `labelEditUiModule` which reacts to the existing `EditLabelAction` and then provides a UI based on the `UIExtension` introduced in Sprotty a while ago. When the user confirms the edit, it sends out a `ApplyLabelEditAction` which can then be processed according to the specific use case. In GLSP we would forward it to the server in order to apply the modification, or you could have a local command just applying it to the `SModel`, or be forwarded to the textual language server.

![direct-edit-ui](https://user-images.githubusercontent.com/588090/56823341-df556280-6853-11e9-8c2c-b2fac1a66209.gif)

Are you, in general, interested to have that in Sprotty directly? As it doesn't put any restrictions on how the label edit is actually be applied (on a server or local), the implementation is rather agnostic of the use case.
The implementation is not perfect yet, but I guess it wouldn't be too much work to brush up the code for Sprotty.

Please let me know what you think.

spoenemann:
Looks good! Yes, I think it makes sense to have the general UI in Sprotty and be able to implement the actual model change in the specific application.

You introduced the Nameable interface some months ago. Could that be used to define a default command for the ApplyLabelEditAction?

planger:
Thanks for the feedback! I'll then go ahead and prepare a PR on that.

> You introduced the Nameable interface some months ago. Could that be used to define a default command for the ApplyLabelEditAction?

Yeah, good idea!

In my initial implementation I directly worked with `EditableLabel`. The `EditLabelAction` is sent out on double-click on an `EditableLabel` (that's already in Sprotty). Then I'd obtain the bounds of the element with the `editableLabel.id` using the introduced `getAbsoluteClientBounds(...)` to compute the bounds for the `UIExtension`. On <kbd>Enter</kbd>, the `ApplyLabelEditAction` is sent out with a label ID. The obvious default command would now just check whether the editable label is an SLabel and apply the new label to `SLabel.text`.

To pick up your idea, a generic command could now instead:
1. `findParentByFeature(editableLabel.id, Nameable)` and set the name of the returned nameable element. This entails a slight risk/limitation that the nameable element must be a direct parent of the editable label (or they are the same element) and you could create an implicit ambiguity by having two editable labels as children of a single nameable element.
[...]

PR #88, opened 2019-05-09
Add user interface for label editing (#83)

planger:
This change introduces a user interface -- as discussed in #83 -- for editing labels.

The UI for label editing is hooked up in the optional `labelEditUiModule` which reacts to the existing `EditLabelAction` and then provides a UI based on the `UIExtension` mechanism introduced a while ago. When the user confirms the edit, the UI extension sends out an `ApplyLabelEditAction`, which should be handled by a command actually updating the model. The default command just updates the text of the `EditableLabel`. In other use cases, such as GLSP or LSP driven diagrams, a different command could be bound that delegates performing the actual update to the respective server.

I tried to leave as much as possible in terms of styling to the CSS. However, the UI extension tries to obtain the label's font size and weight to provide a nice UX while editing. This can be overwritten however in `EditLabelUI.applyFontStyling(...)`.

By default, the editing is initiated by double-clicking an element (as before), or when pressing <kbd>F2</kbd> on a selected element. The target of the double-click or when <kbd>F2</kbd> is pressed can either be an `EditableLabel` directly or an `WithEditableLabel` that points to an `EditableLabel`. The latter is useful when the selectable element is in fact not a label directly but contains one somewhere (sort of the main label of an element).

In #83 we discussed how the default `ApplyLabelEditCommand` should be implemented and @spoenemann suggested to perform the label edit based on the nameable feature. In a first shot, however, it felt more flexible to just directly apply the label edit by updating the `EditableLabel.text` property. Together with the `WithEditableLabel` interface this makes it quite easy and flexible, imho.
However, if you prefer to have the default `ApplyLabelEditCommand` look up an element with the `Nameable` extension, I'm happy to change the implementation.

Moreover, note how the UI extension, command, etc. are exclusively built on `EditableLabel` and `WithEditableLabel`, as well as on `SModelElement` (for obtaining the ID). Thus, the editable label does not even have to be a label, which I think could be useful in some scenarios.

This change also introduces a basic label edit validation mechanism, if an `IEditLabelValidator` is bound. The validation result is added in terms of classes and attributes to the container element of the HTML input.

Finally, this change also slightly adapts the `UIExtension` mechanism so that we can more easily pass context elements into the extension, without relying on attributes that are set in the `SModel` only (such as the selected flag, which we used before).

All functionality is integrated in Sprotty's class diagram example. Here is the obligatory animated gif :)

label-edit2

Please let me know what you think! Thanks in advance!

### DIFF
--- a/examples/classdiagram/class-diagram.html
+++ b/examples/classdiagram/class-diagram.html
@@ -5,6 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>sprotty Class Diagram Example</title>
     <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.2.1/css/bootstrap.min.css">    
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/balloon-css/0.5.0/balloon.min.css">
     <link rel="stylesheet" href="css/page.css">
     <!-- support Microsoft browsers -->
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/dom4/3.0.0/dom4.js">
--- a/examples/classdiagram/src/label-validator.ts
+++ b/examples/classdiagram/src/label-validator.ts
@@ -18,20 +18,20 @@ import { injectable } from "inversify";
 
 @injectable()
 export class ClassDiagramLabelValidator implements IEditLabelValidator {
-    validate(value: string, label: EditableLabel & SModelElement): Promise<EditLabelValidationResult> {
+    async validate(value: string, label: EditableLabel & SModelElement): Promise<EditLabelValidationResult> {
         if (value.length < 1) {
-            return Promise.resolve({
+            return {
                 severity: <Severity>'error',
                 message: 'Name must not be empty'
-            });
+            };
         } else if (value.indexOf('!') !== -1) {
-            return Promise.resolve({
+            return {
                 severity: <Severity>'warning',
                 message: 'Name should not contain exclamation marks'
-            });
+            };
         }
-        return Promise.resolve({
+        return {
             severity: <Severity>'ok', message: undefined
-        });
+        };
     }
 }
--- a/examples/classdiagram/src/label-validator.ts
+++ b/examples/classdiagram/src/label-validator.ts
@@ -1,0 +1,37 @@
+/********************************************************************************
+ * Copyright (c) 2019 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { IEditLabelValidator, EditableLabel, SModelElement, EditLabelValidationResult, Severity } from "../../../src";
+import { injectable } from "inversify";
+
+@injectable()
+export class ClassDiagramLabelValidator implements IEditLabelValidator {
+    validate(value: string, label: EditableLabel & SModelElement): Promise<EditLabelValidationResult> {
+        if (value.length < 1) {
+            return Promise.resolve({
+                severity: <Severity>'error',
+                message: 'Name must not be empty'
+            });
+        } else if (value.indexOf('!') !== -1) {
+            return Promise.resolve({
+                severity: <Severity>'warning',
+                message: 'Name should not contain exclamation marks'
+            });
+        }
+        return Promise.resolve({
+            severity: <Severity>'ok', message: undefined
+        });
+    }
+}
--- a/src/features/edit/edit-label-ui.ts
+++ b/src/features/edit/edit-label-ui.ts
@@ -14,8 +14,10 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 import { inject, injectable, optional } from "inversify";
+import { Action } from "../../base/actions/action";
 import { IActionDispatcherProvider } from "../../base/actions/action-dispatcher";
-import { CommandExecutionContext, CommandResult, SystemCommand } from "../../base/commands/command";
+import { ActionHandlerRegistry, IActionHandler, IActionHandlerInitializer } from "../../base/actions/action-handler";
+import { ICommand } from "../../base/commands/command";
 import { SModelElement, SModelRoot } from "../../base/model/smodel";
 import { TYPES } from "../../base/types";
 import { AbstractUIExtension } from "../../base/ui-extensions/ui-extension";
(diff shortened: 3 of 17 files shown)


planger:
Thanks for your feedback!

> Cool! A few remarks:
> 
> * I would expect the default implementation to block the label change if there are validation errors. That should be customizable of course.

Right, that's missing and should be added in an overrideable method in `EditLabelUI`. I'll append a commit with that at latest end of this week.

> * When I change a class name and the new name is larger than the previous name, it becomes difficult to read. Maybe add a white (or half-transparent) background to the label while editing?

It'd be easy to set a semi-transparent background for the input in the CSS. However, since the input is wider than the label, it wouldn't look very nice because the input would visibly reach beyond the label.

Wouldn't it be better to recompute the micro-layout already during editing to make room for the label being edited, when it becomes wider? This is related to your next comment. Maybe that's a future enhancement, but I think this would give us a very nice UX.

> * We need to recompute the layout after a label has been changed. This applies both to node labels and to edge labels. One way to realize that might be to change the label in the model source (the _external model_) instead of a command (the _internal model_). Maybe this needs further discussion (CC @JanKoehnlein).

In my first attempt I was a bit reluctant to tamper with the external model since the model source is something that might be very different from use case to use case. But I'm happy to look into that. I'll probably wait for the opinion @JanKoehnlein before doing that though to avoid running into the wrong direction.

I think, it would be ideal if we could update the label in the smodel (internal and if necessary also external) also DURING the editing, and recompute the "micro-layout" on the fly. Once this is done, we could adjust the position and size of the HTML input based on the new label bounds. When the user hits enter or leaves the input, we would "commit" the change, otherwise, we just set the old value and recompute layout again either way. This would also solve the issue of the background (your comment above) as well as the issue of the fixed width of the input (currently label-width + 200, which is sort of ugly). Of course, this may have a negative impact on performance, if we recompute the layout often, but I'm not sure how fast this is and whether we can only recompute a certain element hierarchy.

src/features/edit/edit-label-ui.ts
```diff
@@ -0,0 +1,230 @@
...
+    static KIND: string = EditLabelAction.KIND;
+    @inject(TYPES.IActionDispatcherProvider) public actionDispatcherProvider: IActionDispatcherProvider;
+
+    constructor(@inject(TYPES.Action) public action: EditLabelAction) {
+        super();
+    }
+
+    execute(context: CommandExecutionContext): CommandResult {
+        this.actionDispatcherProvider().then(dispatcher =>
+            dispatcher.dispatch(new SetUIExtensionVisibilityAction(EditLabelUI.ID, true, [this.action.labelId])));
```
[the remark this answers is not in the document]
planger: You're right, that would be sufficient. I'll address that in a follow-up commit.

src/features/edit/edit-label-ui.ts
```diff
@@ -0,0 +1,230 @@
...
+                .catch((reason) => this.logger.error(this, 'Error validating edited label', reason));
+        }
+    }
+
+    protected showValidationResult(result: EditLabelValidationResult) {
+        this.clearValidationResult();
+        if (result.message) {
+            this.containerElement.setAttribute('data-balloon', result.message);
+            this.containerElement.setAttribute('data-balloon-pos', 'up-left');
+            this.containerElement.setAttribute('data-balloon-visible', 'true');
```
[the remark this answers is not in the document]
planger: Yes, sort of... they are related to [ballon.css](https://kazzkiq.github.io/balloon.css/), which we load in the class diagram example. Even though it is "just" some HTML attributes that we essentially add, you're probably right that we should extract that into application-specific code. Should we allow to inject a new type eg named `LabelEditValidationUi` which is invoked here and that is implemented in the application-specific code so that it can reuse existing UI frameworks?

src/features/edit/edit-label.ts
```diff
@@ -14,32 +14,131 @@
...
 export interface EditableLabel extends SModelExtension {
+    text: string;
 }
 
 export function isEditableLabel<T extends SModelElement>(element: T): element is T & EditableLabel {
-    return element instanceof SLabel && element.hasFeature(editLabelFeature);
+    return 'text' in element && element.hasFeature(editLabelFeature);
+}
+
+export const withEditLabelFeature = Symbol('withEditLabelFeature');
```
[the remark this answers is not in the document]
planger: Ok, no problem, I'll handle that in a follow-up commit.

examples/classdiagram/class-diagram.html
```diff
@@ -5,6 +5,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>sprotty Class Diagram Example</title>
     <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.2.1/css/bootstrap.min.css">    
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/balloon-css/0.5.0/balloon.min.css">
```
JanKoehnlein: Do we need a CQ for that?

examples/classdiagram/src/label-validator.ts
```diff
@@ -0,0 +1,37 @@
...
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+import { IEditLabelValidator, EditableLabel, SModelElement, EditLabelValidationResult, Severity } from "../../../src";
+import { injectable } from "inversify";
+
+@injectable()
+export class ClassDiagramLabelValidator implements IEditLabelValidator {
+    validate(value: string, label: EditableLabel & SModelElement): Promise<EditLabelValidationResult> {
```
JanKoehnlein: async?

JanKoehnlein:
As for the external model update: We could use a similar approach as with move: Commands are aggregated on the command stack and a `CommitModelAction` is fired when the mouse key is released. The latter is usually picked up by the model source which can update the external model etc. 

Some kind of animation on undo/redo would be nice as well.

The most important aspect for me is that it can integrate with an LSP based server, e.g. validate the name on the server and trigger rename after the label is edited. Seems to be possible with the current approach.

examples/classdiagram/class-diagram.html
planger: Hm, I'm really not very firm on these rules, especially when it comes to "just" referencing libraries... :/ Should I ask the Eclipse IP team?

examples/classdiagram/src/label-validator.ts
planger: Good idea, I fixed that in new commits on this branch.

planger:
Thanks again for all your feedback!

I pushed two new commits that address your comments, except for the following:

> Do we need a CQ for that?

I left this open, because I'm not sure, unfortunately. If you want me to ask the IP team, I'm happy to send them a note.

> As for the external model update: We could use a similar approach as with move: Commands are aggregated on the command stack and a `CommitModelAction` is fired when the mouse key is released. The latter is usually picked up by the model source which can update the external model etc.

Thanks for the hint! For now I only issue a `CommitModelAction` when the label is applied and update the model in the class diagram's model source (to have an effect). If you are fine with that, I'd rather work on an improved UX (ie. updating the model while editing and adapting the input size) in a subsequent PR.

I hope the rest is addressed accordingly:
  - Prevent applying label edit if label validation resulted in an error
  - Issue a CommitModelAction and let class diagram example update model
  - Move symbols and SModelExtensions into model.ts
  - Replace command for initiating label edit with action handler
  - Add semi-transparent background for label input for now
  - Extract decoration of label edit validation result into client application

> The most important aspect for me is that it can integrate with an LSP based server, e.g. validate the name on the server and trigger rename after the label is edited. Seems to be possible with the current approach.

@JanKoehnlein Yeah, that should be easily possible. I'm doing the same with the GLSP server already and didn't have any troubles integrating it there. If you run into issues for LSP, please let me know.

Are you fine with merging these additional changes? Thanks!

spoenemann:
>> Do we need a CQ for that?

> I left this open, because I'm not sure, unfortunately. If you want me to ask the IP team, I'm happy to send them a note.

I don't think that's necessary since we use it only in one of the examples, not in the library itself (so it's not a _prerequisite_ of Sprotty, see below). The same applies to Bootstrap and other stuff we use in the examples.

https://www.eclipse.org/projects/handbook/#ip-third-party

PR #98, opened 2019-08-20
Enhance context information for command palette action providers

planger:
This additional info includes the last known mouse position on the
diagram as well as the text entered into the command palette so far.
This additional info is helpful to allow for more sophisticated
and context sensitive suggestions and ordering of those suggestions.

Moreover, this change makes it easier to override the behavior for
fetching the command palette actions on typing, introduces a debounce
delay to avoid flooding the action providers with requests on each
keystroke, and finally, it adds a HTML class to the input element
while loading actions to allow visually showing that actions are
currently being loaded.

### DIFF
--- a/src/base/views/mouse-tool.ts
+++ b/src/base/views/mouse-tool.ts
@@ -23,8 +23,6 @@ import { TYPES } from "../types";
 import { DOMHelper } from "./dom-helper";
 import { IVNodeDecorator } from "./vnode-decorators";
 import { on } from "./vnode-utils";
-import { findParentByFeature } from "../model/smodel-utils";
-import { isViewport } from "../../features/viewport/model";
 import { Point } from "../../utils/geometry";
 
 @injectable()
@@ -33,7 +31,7 @@ export class MouseTool implements IVNodeDecorator {
     @inject(TYPES.IActionDispatcher) protected actionDispatcher: IActionDispatcher;
     @inject(TYPES.DOMHelper) protected domHelper: DOMHelper;
 
-    constructor(@multiInject(TYPES.MouseListener)@optional() protected mouseListeners: MouseListener[] = []) {}
+    constructor(@multiInject(TYPES.MouseListener) @optional() protected mouseListeners: MouseListener[] = []) { }
 
     register(mouseListener: MouseListener) {
         this.mouseListeners.push(mouseListener);
@@ -45,7 +43,7 @@ export class MouseTool implements IVNodeDecorator {
             this.mouseListeners.splice(index, 1);
     }
 
-    protected getTargetElement(model: SModelRoot, event: MouseEvent): SModelElement |undefined {
+    protected getTargetElement(model: SModelRoot, event: MouseEvent): SModelElement | undefined {
         let target = event.target as Element;
         const index = model.index;
         while (target) {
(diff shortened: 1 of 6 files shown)


src/base/views/mouse-tool.ts
```diff
@@ -197,3 +200,24 @@ export class MouseListener {
...
     }
 }
+
+@injectable()
+export class MousePositionTracker extends MouseListener {
+
+    protected lastPosition: Point | undefined;
+
+    mouseMove(target: SModelElement, event: MouseEvent): (Action | Promise<Action>)[] {
+        const viewport = findParentByFeature(target, isViewport);
```
spoenemann: I think this should be generalized using `localToParent`, maybe we need another utility function along the lines of `getAbsoluteBounds` and `getAbsoluteClientBounds` from `features/bounds/model.ts`?

planger:
Thanks for the feedback!

> Can the effects of this change be seen in the built-in examples?

Only partially, since our example `RevealNamedElementActionProvider` doesn't use the additional information (mouse position and entered text) that we've introduced with this PR and it provides its actions immediately as the computation is local. For actions that are fetched from a server, the text input field now gets a CSS class assigned while waiting for the turnaround with the server, so that we can show a visual feedback while waiting.

The remaining improvements are visible in the class diagram example (highlighting of the matching text, icon support, you can also observe the debounce delay when typing):

image

The other changes mainly concern extensibility by extracting certain behavior that clients may like to customize into overrideable methods (such as `CommandPalette.updateAutoCompleteActions(...)`, `CommandPalette.renderLabeledActionSuggestion(...)`, `CommandPalette.renderIcon(...)`).

> I think this should be generalized using `localToParent`, maybe we need another utility function along the lines of `getAbsoluteBounds` and `getAbsoluteClientBounds` from `features/bounds/model.ts`?

I'm happy to extract this into a generic utility function. However, I'm not sure localToParent is helpful here, but I might miss something. What we would like to have is a diagram-relative position of the mouse (because this position makes sense to every component that knows the diagram, including a server). The mouse event is fired by the div containing the diagram, so the offset of the event seems to be the right choice. The only thing we need to take into account to get the diagram-relative position is the zoom level and the scroll. Therefore, we obtain the viewport.
How do you suggest to use localToParent?

Thanks again!

spoenemann:
> How do you suggest to use localToParent?

It should be `parentToLocal`, sorry. If you look at the `parentToLocal` implementation of `ViewportRootElement`, it does exactly the same calculation that you have in your MousePositionTracker, so you should call that method on the root element instead.

MousePositionTracker should clearly define to which element the reported positions are relative to. The diagram may contain multiple model elements that define a local coordinate system, so this question is not trivial.

planger:
> > How do you suggest to use localToParent?
> 
> It should be `parentToLocal`, sorry. If you look at the `parentToLocal` implementation of `ViewportRootElement`, it does exactly the same calculation that you have in your MousePositionTracker, so you should call that method on the root element instead.
> 
> MousePositionTracker should clearly define to which element the reported positions are relative to. The diagram may contain multiple model elements that define a local coordinate system, so this question is not trivial.

Ah, I see. Thanks for the clarification! I obviously missed that this computation already exists. a3e1bf1 now reuses `ViewportRootElement.parentToLocal` and adds a comment making it more clear that this is the last mouse position relative to the diagram root.

JanKoehnlein:
Does anybody know whether we need a CQ for using font awesome this way?

planger:
> Does anybody know whether we need a CQ for using font awesome this way?

No, I don't think so. @spoenemann researched that and commented the result here:
https://github.com/eclipse/sprotty/pull/88#issuecomment-496109019
 > https://www.eclipse.org/projects/handbook/#ip-third-party

PR #103, opened 2019-08-30
Implemented request/response actions, code cleanup

### LINKED ISSUE #8: Support request / response scheme
From theia-ide/sprotty#218:

> We have several actions that appear in request / response pairs: 
> 
>  * RequestModelAction / SetModelAction
>  * RequestPopupModelAction / SetPopupModelAction
>  * RequestBoundsAction / SetBoundsAction
>  * RequestExportSvgAction / ExportSvgAction
> 
> We should add a mechanism that allows to match a response to a corresponding request. This is necessary in case multiple actions of the same kind are fired at the same time.

### LINKED ISSUE #85: [arch] reading SModel from the outside
When using sprotty in an UI framework like Theia, some user interactions are triggered outside of the sprotty architecture, e.g. keystrokes or drag and drop. You sometimes need to access certain properties of the SModel in order to turn these into sprotty actions, e.g.

- transform a screen position into a diagram coordinate (needs Viewport)
- access the selection

While within sprotty, event listeners can access the SModel, these external listeners can't.

For other cases, such as bounds, we often process certain action in the model source and keep the external model updated. External listeners can then get the current values from the model source. Unfortunately, this doesn't seem to be a viable solution for viewport actions such as `center`, as the necessary calculations and the update of the viewport happen directly in the respective command.

As a second solution, on can implement some kind of pre-action and a pre-command that gathers the necessary information and issues the real action. This sounds a bit complicated, but is safer as we don't make assumptions which actions change the needed properties.

A third solution would be to implement services, like a selection or viewport provider, that get the values directly from the current model in the command stack. This kind of breaks the encapsulation we had.

spoenemann:
Fixes #8, fixes #85.

 * Added `RequestAction` and `ResponseAction` interfaces.
 * Added a `request` method to `IActionDispatcher` to dispatch requests and wait for the response.
 * Split `Viewer` in three classes `ModelViewer`, `HiddenModelViewer`, `PopupModelViewer`.
 * Allow commands to return a `cause` action.
 * Pass the `cause` action to the viewer and its decorators so they can link any created response action with the original request.
 * Added `ModelRequestCommand` to ease the implementation of requests for model details.
 * Added `getSelection` and `getViewport` to `LocalModelSource` based on corresponding request/response actions.


src/base/commands/command.ts
```diff
@@ -52,14 +52,22 @@ export interface ICommand {
...
-export type CommandResult = SModelRoot | Promise<SModelRoot>;
+export type CommandResult = SModelRoot | Promise<SModelRoot> | DetailedCommandResult;
+
+/**
+ * The `DetailedCommandResult` allows to specify whether the model has changed
+ * and the original action that caused the command to be executed. In case such
+ * an action is given, it is passed to the viewer in order to link any
+ * subsequent response action to the original request.
+ */
+export type DetailedCommandResult = { model: SModelRoot, isChanged: boolean, cause?: Action };
```
planger: I'm wondering whether we should rename `isChanged` to `isModelChanged` as it might be unclear in the context of a command what `isChanged` means.

spoenemann: Renaming `CommandResult` to `CommandReturn` and `DetailedCommandResult` to `CommandResult`. We may further extend it in the future. Currently we need the `cause` only for request actions, but there is no need to restrict it here since the actual action kind must be checked in the decorators anyway.

src/features/bounds/bounds-manipulation.ts
```diff
@@ -39,10 +40,16 @@ export class SetBoundsAction implements Action {
...
+    static readonly KIND: string  = 'requestBounds';
+    readonly kind = RequestBoundsAction.KIND;
 
-    constructor(public readonly newRoot: SModelRootSchema) {
+    constructor(public readonly newRoot: SModelRootSchema,
+                public readonly requestId: string = '') {}
+
+    /** Factory function to dispatch a request with the `IActionDispatcher` */
+    static create(newRoot: SModelRootSchema): RequestAction<ComputedBoundsAction> {
+        return new RequestBoundsAction(newRoot, generateRequestId());
```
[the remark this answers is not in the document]
spoenemann: Currently we mark all properties of actions as `readonly`. Adding the request id after action creation would break that contract.

We need such a static function anyway for each request action because the TypeScript type system is currently unable to match the correct type parameter for a class implementing a parameterized interface (it does work in Java).

PR #117, opened 2019-09-13
Added ActionHandlerRegistration

### LINKED ISSUE #48: [di] Consider creating action handlers using DI
#46 introduces dependency injection for commands and views. We should consider applying the same approach to action handlers.

spoenemann:
Fixes #48. The new `ActionHandlerRegistration` is the now preferred alternative to an `IActionHandlerInitializer`. The latter is still supported and used for `ModelSource` as well as `CommandActionHandlerInitializer`. I replaced other usages of the old initializer interface with the new registration.

What do you think?


spoenemann:
No, it's additional to the existing API.

PR #119, opened 2019-09-16
[bounds] only set size of server-side layouted elements

### LINKED ISSUE #118: [bounds] ComputedBoundsAction forces positions even if not responsible
A `ComputedBoundsAction` transfers the sizes and positions of elements which have been measured from the hidden DOM. For elements that are not subject to client layout, we only need their size, but we force a position of (0,0) by using double numbers in `ElementAndBounds`. This overwrites a positions that may be calculated on the server, e.g. extracted from the semantic model.

For server-side layouted elements (i.e. children of a container that doesn't have a layout) we should not set a position in `ComputedBoundsAction`.

JanKoehnlein:
Fixes #118


JanKoehnlein:
@spoenemann I've added on more commit. Could you review again?

PR #136, opened 2019-10-15
Fixed zooming for Firefox

### LINKED ISSUE #114: Zooming is strange on Firefox / Windows
Mouse wheel zooming has a strange behavior in Firefox (tested on Windows).
 * It is slow: I have to turn the wheel a lot to make some zooming happen.
 * When I hold the mouse cursor over the diagram background, it zooms as expected. But when I hold it over a node, it seems to zoom towards the upper left corner of the viewport.

spoenemann:
Fixes #114.

 * Take into account the `deltaMode` of the WheelEvent.
 * Use `clientX` / `clientY` instead of the unreliable `offsetX` / `offsetY`


spoenemann:
Just tried it, it works well even on Edge. The zooming is quite coarse there, though. Probably we need to fine-tune the numeric constants.

PR #138, opened 2019-10-28
Allowed the CenterAction to retain the zoom level

tivervac:
When centering, you don't always want the zoom level to be reset.

Needs [this](https://github.com/eclipse/sprotty-server/pull/58) as well

### DIFF
--- a/src/features/viewport/center-fit.ts
+++ b/src/features/viewport/center-fit.ts
@@ -30,7 +30,8 @@ import { TYPES } from "../../base/types";
 
 /**
  * Triggered when the user requests the viewer to center on the current model. The resulting
- * CenterCommand changes the scroll setting of the viewport accordingly and resets the zoom to its default.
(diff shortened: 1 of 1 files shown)


src/features/viewport/center-fit.ts
```diff
@@ -157,9 +159,11 @@ export abstract class BoundsAwareViewportCommand extends Command {
 
 export class CenterCommand extends BoundsAwareViewportCommand {
     static readonly KIND = CenterAction.KIND;
+    retainZoom: boolean;
```
spoenemann: You don't need to add a property for `retainZoom` because the `action` is already stored as protected property in the constructor (this is indicated by the `protected` modifier of the constructor argument).

tivervac: Done, thanks

JanKoehnlein:
Merged. Thanks for the patch!

PR #141, opened 2019-11-02
Support for cycling through command palettes (#140)

### LINKED ISSUE #140: Support cycling through command palette groups
As known from many IDEs that support a command palette, many allow to cycle through multiple groups of palettes by hitting <kbd>Ctrl+Space</kbd> again and again. It would be useful to have a similar support for Sprotty's command palette as well.
Since this probably won't be needed by all users of the command palette, this should be imho an optional feature that only mildly affects the API.

planger:
Fixes #140 
For illustration purposes, I integrated this -- as usual -- into the class diagram example (below I pushed <kbd>Ctrl+Space</kbd> multiple times).
Peek 2019-11-02 20-50

### DIFF
--- a/src/features/command-palette/action-providers.ts
+++ b/src/features/command-palette/action-providers.ts
@@ -20,12 +20,12 @@ import { TYPES } from "../../base/types";
 import { toArray } from "../../utils/iterable";
 import { ILogger } from "../../utils/logging";
 import { isNameable, name } from "../nameable/model";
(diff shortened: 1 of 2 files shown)


planger:
While I'm at it, this PR also clears the command palette content before it shows it. This is imho the better default behavior.

planger:
Thanks @JanKoehnlein!

PR #142, opened 2019-11-05
[view] use tabindex 0 for svg element

JanKoehnlein:
Use tabindex 0 (automatically add it to the sequential tab order) for SVG element.
Tabindex has to be set in order to allow focus on FireFox


JanKoehnlein:
@planger Thanks for reviewing

PR #145, opened 2019-11-06
Append element subtype as view class

### LINKED ISSUE #137: Add subtype as class to view
For some views (e.g. `SLabelView`, `SCompartmentView`) the subtype is added as a class. However for other views such as `RectangularNodeView` this is not added.

Having the subtype available makes custom styling a lot easier, so it would be handy if all views had this functionality.

tomvdbussche:
Fixes #137


tomvdbussche:
> LGTM

Great! I don't think I have the power to merge this request so if someone could do this for me that would be great :smile:

JanKoehnlein:
Merged. Thanks for the patch!

PR #148, opened 2019-11-16
Improve loading indicator of command palette

planger:
This change simplifies showing a loading indicator while the
command palette is loading actions, e.g. with fontawesome.
It is now easy to just override loadingIndicatorClasses, which
will be assigned to a span element after the command palette
input. With CSS and fontawesome, clients can then show a
spinning icon while the actions are loading with plain css.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,6 +6,13 @@ spec:
   - name: node
     image: node:8.12-alpine
     tty: true
+    resources:
+      limits:
+        memory: "2Gi"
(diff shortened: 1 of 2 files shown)


planger:
I had to increase the resources in the CI with ed3ccb1 to get the build through. If you prefer, I can also put that the change to the Jenkinsfile into an own PR.

planger:
@JanKoehnlein Thanks for the review! :+1:

PR #151, opened 2019-12-02
Avoid adding duplicate loading indicators in command palette

planger:
If for some reason the update of items fails and we never get to removing
the loading indicator, we have to avoid re-adding another one the
next time we show the command palette. Otherwise we may end up with
a remaining loading indicator that is never removed.
Also this change makes the onSelect customizable by subclasses.


planger:
@JanKoehnlein or @spoenemann is it ok with you for me to merge this change? I'd need this change in the Sprotty next builds rather soon, to be able to consume it in a project. Sorry for poking... and thanks a lot in advance!

planger:
Thanks, @spoenemann!

PR #153, opened 2019-12-09
Reset previous hover feedback on mouseover

### LINKED ISSUE #152: Hover feedback is not removed if there is no mouseout event for that element
It may happen that no mouseout event is dispatched resulting in a hover feedback to remain active even though other hover feedbacks are already created.

An instance of that is easy to reproduce when rendering HTML over a Sprotty diagram (popup, context menu, command palette, ...):
If the user exits the hovered element via such HTML elements and re-enters the diagram outside of the originally hovered element, the hover feedback of the originally hovered element is never removed until we cross it again with the mouse and produce another mouseout event.

hover-feedback-bug

planger:
With this change, I suggest to always reset the hover feedback of the previous element on mouseover of any diagram element. This ensures that we reset the hover feedback also if the user exited the element without producing any mouseout event of the original hovered element.

Resolves #152

### DIFF
--- a/src/features/hover/hover.spec.ts
+++ b/src/features/hover/hover.spec.ts
@@ -65,6 +65,11 @@ describe('hover', () => {
     class HoverableTarget extends SModelElement implements Hoverable {
         hoverFeedback: boolean = false;
 
+        constructor(id: string = "1") {
+            super();
+            this.id = id;
+        }
+
         hasFeature(feature: symbol): boolean {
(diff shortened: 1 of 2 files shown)


planger:
> There's a failing test for which the expectation needs to be updated. Otherwise I think it's good.

Thanks, Miro! I've force-pushed a change that resets the state of the hover listener after the new test so that the subsequent test isn't affected anymore.

Issue #162, opened 2020-02-10
Some viewport related features are not working unter IE11


tortmayr:
This is caused by `window.scrollX`/window.scrollY` which are not supported by IE11. See also (https://stackoverflow.com/questions/44757869/window-pageyoffset-vs-window-scrolly-on-ie11)

I believe that  IE 11 is still the only  browser back-end that works in SWT's browser widget on windows so the sprotty code base should still be IE11 compatible.

PR #163, opened 2020-02-10
#162: IE11-compatible window scroll retrieval

tortmayr:
Use of window.pageXOffset instead of window.scrollX  (same for Y) to support all browsers.
https://github.com/eclipse/sprotty/wiki/Browsers-and-Compatibility was updated accordingly


src/features/bounds/model.ts
```diff
@@ -117,8 +117,8 @@ export function getAbsoluteClientBounds(element: SModelElement, domHelper: DOMHe
     const svgElement = document.getElementById(svgElementId);
     if (svgElement) {
         const rect = svgElement.getBoundingClientRect();
-        x = rect.left + window.scrollX;
-        y = rect.top + window.scrollY;
+        x = rect.left + window.pageXOffset;
+        y = rect.top + window.pageYOffset;
```
spoenemann: Shall we use `getWindowScroll()` here, too?

tortmayr: Done

Issue #169, opened 2020-03-18
Text wrapping support


planger:
We'd like to have support for text wrapping within shapes (e.g. for comments in a diagram or long labels, etc.) and I think it could make sense to add support for text wrapping to Sprotty. But before we start working on this, I wanted to get your feedback on whether that's something we'd like to have in Sprotty at all and get your opinion which approach may be the best to implement that.

The reason this is not straightforward is because SVG itself doesn't really support text wrapping (maybe SVG 2.0 will eventually provide `textArea` at some point but this doesn't help us now). So to address this in Sprotty, we may need a bit more than. After a first round of research and I think we have the following options. Please let me know what you think and if you would agree to have this in Sprotty after all.

1. Use `foreignObject` to embed HTML in the SVG. This would be quite straightforward and would have the benefit that we could directly support also HTML-based "rich-text" without any additional work. The drawback is that IE11 doesn't support `foreignObject` (Chromium and Firefox do though). Also, there may be some glitches when it comes to selection, focus and click behavior. I'm not sure how browsers would behave when forwarding the mouse events from the foreign object to the underlying SVG, but I guess that's something that can be addressed.
2. Use a third-party library, such as `d3plus-text`. Essentially they dynamically split the text into separate SVG `text` and `textSpan` elements and compute their positions according to a text flow with a given width. However, this would pull in a dependency to Sprotty and require some dedicated implementations in Sprotty to take the wrapping into account for (micro-)layouting.
3. Implement something similar to (2) in Sprotty directly with a dedicated SModel label type (splitting text into `textSpan`s), dedicated `Layouter` (to compute the height of an element which is affected by the text and its wrapping according to some specific layout options) and a `IVNodeDecorator` which positions elements according to the text flow. The layouter knows the size of the wrappable words based on the hidden SModel and somehow has to preserve this information so that the IVNodeDecorator can then compute the wrapping and apply the positioning. The reason I'd like to put the computation and positioning into the node decorator is to enable live wrapping while users e.g. change the shape size. Only doing that in the layouter, we'll just be able to rewrap after a model update. This approach would be Sprotty- and SVG-native, but would also be more work and more code that we'd have to maintain for very basic functionality (word-wrapping).
4. Use UIExtension to overlay HTML with text wrapping on top of the diagram. However, this may cause some focus issues, probably worse than approach (1), with the benefit of IE support, but the drawback that we'd need an additional workaround for the SVG export.

[...]

PR #171, opened 2020-03-24
Add support for foreignObject and enhance label editing with multi-line support

planger:
https://github.com/eclipse/sprotty/issues/169

### DIFF
--- a/src/features/edit/edit-label-ui.ts
+++ b/src/features/edit/edit-label-ui.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 EclipseSource and others.
+ * Copyright (c) 2019-2020 EclipseSource and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -25,7 +25,7 @@ import { SetUIExtensionVisibilityAction } from "../../base/ui-extensions/ui-exte
 import { DOMHelper } from "../../base/views/dom-helper";
 import { ViewerOptions } from "../../base/views/viewer-options";
 import { CommitModelAction } from "../../model-source/commit-model";
-import { matchesKeystroke } from "../../utils/keyboard";
+import { matchesKeystroke, KeyCode, KeyboardModifier } from "../../utils/keyboard";
 import { getAbsoluteClientBounds } from "../bounds/model";
 import { getZoom } from "../viewport/zoom";
 import {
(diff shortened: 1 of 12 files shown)


planger:
Peek 2020-03-24 09-36

src/lib/generic-views.tsx
```diff
@@ -35,3 +38,26 @@ export class PreRenderedView implements IView {
...
+        const node = <g>
+            <foreignObject requiredFeatures='http://www.w3.org/TR/SVG11/feature#Extensibility'
+                height={model.bounds.height} width={model.bounds.width} x={0} y={0}>
+                {foreignObjectContents}
+            </foreignObject>
+            {context.renderChildren(model)}
+        </g>;
+        setAttr(node, 'class', model.type);
+        const subType = getSubType(model);
+        if (subType) setAttr(node, 'class', subType);
```
tomvdbussche: The subtype is added for every view in `CssClassPostprocessor` so it's not be needed here.

src/features/edit/edit-label-ui.ts
```diff
@@ -193,24 +210,33 @@ export class EditLabelUI extends AbstractUIExtension {
...
+            const zoom = getZoom(this.label);
             const bounds = getAbsoluteClientBounds(this.label, this.domHelper, this.viewerOptions);
-            x = bounds.x;
-            y = bounds.y;
-            height = bounds.height;
-            width = bounds.width + (this.additionalInputWidth * getZoom(this.label));
+            x = bounds.x + (this.label.editControlBoundsCorrection ? this.label.editControlBoundsCorrection.x * zoom : 0);
+            y = bounds.y + (this.label.editControlBoundsCorrection ? this.label.editControlBoundsCorrection.y * zoom : 0);
+            height = bounds.height + (this.label.editControlBoundsCorrection ? this.label.editControlBoundsCorrection.height * zoom : 0);
+            width = bounds.width + (this.label.editControlBoundsCorrection ? this.label.editControlBoundsCorrection.width * zoom : 0);
```
tomvdbussche: These changes have some unintended side-effects for vertical editable labels (see edges in the class diagram example)

planger: Thanks, good catch! This is actually a tough one...
Previously, this just didn't cause problems "by accident" because we added a quite generous additional width and didn't apply the height to the input but only to the container (change below `this.editControl.style.height ...`).
We don't really know whether the label is rotated here. I guess the most pragmatic approach is to set a fixed height and width (as before this change) and let the label implementation decide if it wants to set height and width in a specific way (e.g. to the bounds). I'll make a suggestion in a new commit very soon.

planger:
Thanks a lot @tomvdbussche for your review! I hope https://github.com/eclipse/sprotty/pull/171/commits/70252d490de07800e4fb90e34d0c74b0c9451e80 addresses your comments.

src/features/edit/edit-label-ui.ts
tomvdbussche: That makes sense

PR #173, opened 2020-04-08
Fix label validation if edit control is text area

planger:
We always add the event listener for label validation to the input element
but not to the text area. For the label edit validation to work, we need to
also add an event listener to the text area (ie the element that is passed
as argument in configureAndAdd(...).


planger:
Thanks @tomvdbussche!

PR #184, opened 2020-07-23
Added ability to reject request actions

spoenemann:
It should be possible to reject request actions. This change introduces RejectAction for this purpose.

This is particularly important for requests that are forwarded to a diagram server, e.g. RequestModelAction. We should implement the new feature in sprotty-server, too, after this is merged.


spoenemann:
I thought about that, too, but what type should we use? `any`?

JanKoehnlein:
Not `any`, as these messages are likely to be serialized, but maybe something like in https://github.com/microsoft/TypeScript/issues/1897#issuecomment-338650717

spoenemann:
I added a `detail` property to the new action and used the proposed JSON types where I found them appropriate. Please review.

PR #186, opened 2020-08-19
Catch unavailable IContextMenuService

### LINKED ISSUE #172: Right-click throws exception
Right-clicking anywhere in the example diagrams prints the following error to the console:

```
uncaught exception: undefined
```

spoenemann:
Fixes #172.

### DIFF
--- a/src/features/context-menu/mouse-listener.ts
+++ b/src/features/context-menu/mouse-listener.ts
@@ -13,9 +13,9 @@
  *
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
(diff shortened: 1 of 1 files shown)


tortmayr:
In general this looks good to me. But wouldn't it be better to address this in the corresponding `di.config.ts`? i.e  only bind the provider if a `IContextMenuService` is bound.

spoenemann:
That's not possible because usually `contextMenuModule` is loaded before the Inversify model that binds the actual IContextMenuService, so the binding is not present yet.

tortmayr (approved):
I see. I have one last remark: The @optional() annotations for the context menu providers  should probably be removed (or the properties should be made actually optional)

PR #187, opened 2020-08-19
Added getParent and getRoot to SModelIndex

spoenemann:
This is a preparation for eclipse/sprotty-layout#12. There are situations where one needs a reference to a parent element, which is not available in SModelElementSchema. The new code does not add any memory overhead for the internal model (SChildElement already has a `parent` property).


spoenemann:
I updated Typescript to version 3.8.3 so we can use conditional types.

JanKoehnlein (approved):
LGTM

PR #202, opened 2021-01-20
Align inversify dependency with Theia

### LINKED ISSUE #201: Incompatible inversify version when used with Theia
Sprotty declares the dependency to inversify with [`5.0.1`](https://github.com/eclipse/sprotty/blob/ae671b54aa48f991ca036906bd04ac0ad309ab9f/package.json#L53) while Theia specifies the same dependency with all minor&patch versions higher included: [`^5.0.1`](https://github.com/eclipse-theia/theia/blob/4a3e133bfdab7076dd0290a59f5523c85673d3f2/packages/core/package.json#L36), which currently resolves to version `5.0.5`. This caused the following incompatibility for me:

```
src/browser/diagram-widget-frontend-module.ts(9,26): error TS2345: Argument of type 'import("/home/philip/Temp/theia-sprotty-view/node_modules/inversify/dts/interfaces/interfaces").interfaces.Bind' is not assignable to parameter of type 'import("/home/philip/Temp/theia-sprotty-view/node_modules/@theia/core/node_modules/inversify/dts/interfaces/interfaces").interfaces.Bind'.
  Call signature return types 'BindingToSyntax<any>' and 'BindingToSyntax<T>' are incompatible.
    The types of 'to(...).inSingletonScope().when' are incompatible between these types.
      Type '(constraint: (request: Request) => boolean) => BindingOnSyntax<any>' is not assignable to type '(constraint: (request: Request) => boolean) => BindingOnSyntax<T>'.
        Types of parameters 'constraint' and 'constraint' are incompatible.
          Types of parameters 'request' and 'request' are incompatible.
            Type 'import("/home/philip/Temp/theia-sprotty-view/node_modules/inversify/dts/interfaces/interfaces").interfaces.Request' is not assignable to type 'import("/home/philip/Temp/theia-sprotty-view/node_modules/@theia/core/node_modules/inversify/dts/interfaces/interfaces").interfaces.Request'.
              The types of 'parentContext.container' are incompatible between these types.
                Type 'Container' is missing the following properties from type 'Container': getAllTagged, getAllNamed
```

It can be worked around with selective resolutions in a project's package.json, but it is a bit ugly to tamper with third-party dependency resolutions.

Any objections to change Sprotty's dependency to `^5.0.1` to be inline with Theia so that it resolves to the same version?

planger:
Fixes #201

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
   "dependencies": {
     "autocompleter": "5.1.0",
     "file-saver": "2.0.2",
-    "inversify": "5.0.1",
+    "inversify": "^5.0.1",
     "snabbdom": "0.7.3",
     "snabbdom-jsx": "0.4.2",
     "snabbdom-virtualize": "0.7.0"


spoenemann (approved):
:+1:

PR #203, opened 2021-02-03
Text element dragging fix for firefox

### LINKED ISSUE #195: Cannot move diagram in Firefox
Hi,

I'm running the [Yangster extension](https://open-vsx.org/extension/typefox/yang-vscode) in Theia. When opening a diagram it renders just fine but moving the diagram is not possible. I try to move by clicking on the diagram and then move the mouse.

Works in Chrome but not using:

Firefox browser 82.0.2 (64-bit )
Windows 10 Enterprise version 1809

Thanks,
Andreas

Darkgaja:
Fixes #195

After the first `mouseDown` event on an text element, firefox always triggers a `mouseEnter` event as well (other browsers don't).
This will undo the state changes made by the `mouseDown` event. 

To fix this, we will simply check on `mouseEnter` whether the element is already in a dragging state before we call `mouseUp`.

### DIFF
--- a/src/features/move/move.ts
+++ b/src/features/move/move.ts
@@ -505,7 +505,7 @@ export class MoveMouseListener extends MouseListener {
     }
 
     mouseEnter(target: SModelElement, event: MouseEvent): Action[] {
-        if (target instanceof SModelRoot && event.buttons === 0)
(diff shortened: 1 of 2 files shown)


spoenemann (approved):
Seems to work, thanks!

PR #204, opened 2021-02-03
Fixing examples redirection issue

### LINKED ISSUE #198: Invalid redirection while "Running the examples"
Following the instructions on page
https://github.com/eclipse/sprotty/wiki/Running-the-examples
(either with master or 0.9.0 tag)
When I go to http://127.0.0.1:8080 with my browser (Firefox, Chrome, even tried with IE), I get an error page.
Looks like there is an invalid redirection. Firefox says the request can't succeed ("La requête ne peut aboutir" in french), Chrome says "ERR_INVALID_REDIRECT", IE tries to go to http://127.0.0.1:8080/////// (with many slashes).

On the other hand, going to http://127.0.0.1:8080/index.html works like a charm.

So we could either :
- try and fix the redirection
- update the wiki to add index.html

This is easy to workaround, but this may easily discourage newcomers trying Sprotty.

Darkgaja:
Fixes #198

Some version of ecstatic (dependency of http-server) introduced redirection issues under windows, as discussed in here https://github.com/http-party/http-server/issues/525.

Updating the dependency resolves this issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "circular-dependency-plugin": "^5.2.0",
     "core-js": "^3.2.1",
     "css-loader": "^3.2.0",
-    "http-server": "^0.11.1",
+    "http-server": "^0.12.3",
     "jenkins-mocha": "^8.0.0",
     "jsdom": "^15.1.1",
     "jsdom-global": "^3.0.2",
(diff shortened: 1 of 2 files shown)


spoenemann (approved):
Thanks for the fix!

PR #205, opened 2021-02-03
Only move element if parent is not moving

### LINKED ISSUE #161: Moving a container and a contained node moves the node twice
I'll explain this on the classdiagram example.

When selecting the example package and the class Bar within the package (or simply everything with Ctrl+A) and then moving the selection, the class Bar is also moved.
I would expect that contained elements are not moved inside the container if the container is also moved.

Darkgaja:
Fixes #161 

While selecting the elements to move, we ignore those whose parents are already selected.
I also took the freedom to apply a formatter on the whole `move.ts` file.

### DIFF
--- a/src/features/move/move.ts
+++ b/src/features/move/move.ts
@@ -46,8 +46,8 @@ export class MoveAction implements Action {
     kind = MoveCommand.KIND;
 
     constructor(public readonly moves: ElementMove[],
-        public readonly animate: boolean = true,
-        public readonly finished: boolean = false) {
+                public readonly animate: boolean = true,
+                public readonly finished: boolean = false) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/features/move/move.ts
```diff
@@ -428,17 +428,30 @@ export class MoveMouseListener extends MouseListener {
...
                 }
-            });
+            }
+        });
+    }
+
+    private isChildOfSelected(selectedElements: Set<SModelElement>, element: SModelElement): boolean {
+        while (element instanceof SChildElement) {
+            element = element.parent;
+            if (selectedElements.has(element)) {
```
spoenemann, line 450:
```suggestion
            if (isMoveable(element) && selectedElements.has(element)) {
```

src/features/move/move.ts
```diff
@@ -46,8 +46,8 @@ export class MoveAction implements Action {
     kind = MoveCommand.KIND;
 
     constructor(public readonly moves: ElementMove[],
-                public readonly animate: boolean = true,
-                public readonly finished: boolean = false) {
```
spoenemann, line 50: IMO the previous formatting was more useful here.

spoenemann (changes requested):
Thanks for the contribution! It works, but it also breaks moving bend points of selected edges. Maybe the suggested change below fixes that (I think that condition should be added anyway).

spoenemann (approved):
:+1:

PR #217, opened 2021-05-12
Redo with CrtlCmd+Y

sorekz:
Consider Ctrl+Y as redo keyboard shortcut.
Browsers usually support both Ctrl+Y and Ctrl+Shift+Z in e.g., input elements.

### DIFF
--- a/src/features/undo-redo/undo-redo.ts
+++ b/src/features/undo-redo/undo-redo.ts
@@ -33,7 +33,7 @@ export class UndoRedoKeyListener extends KeyListener {
     keyDown(element: SModelElement, event: KeyboardEvent): Action[] {
         if (matchesKeystroke(event, 'KeyZ', 'ctrlCmd'))
             return [new UndoAction];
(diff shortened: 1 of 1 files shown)


spoenemann:
This seems to be OS-specific. On Mac OS, `Cmd+Y` opens the browser history.

What OS are you using?

sorekz:
I'm using Windows. There `Ctrl+Z` / `Ctrl+Y` are the key combinations for Undo / Redo. Before sprotty I didn't even know you can use `Ctrl+Shift+Z`, too. The shortcut to open the browser history on Windows is `Ctrl+H`.
Just checked my Ubuntu and both also work there but the menu there says `Ctrl+Shift+Z` for Redo

spoenemann:
Then I'd propose that the additional combination should not be used for Mac OS. You can use this utility function to detect the OS.

spoenemann (approved):
Thanks!

Issue #218, opened 2021-05-19
Issue with Angular 11 because of Typscript 4


BjBe82:
We are currently trying to use Sprotty in an Angular 11 Application which requires to use Typescript 4+.
As it looks like there are changes in Typescript 4 which causes errors in Sprotty because some “issues” are now reported as errors.

- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#properties-overriding-accessors-and-vice-versa-is-an-error
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#operands-for-delete-must-be-optional

```
src/base/model/smodel.ts:113:16 - error TS2790: The operand of a 'delete' operator must be optional.
113         delete (child as {parent: SParentElement}).parent;
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/model/smodel.ts:123:28 - error TS2790: The operand of a 'delete' operator must be optional.
123                     delete (child as {parent: SParentElement}).parent;
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/model/smodel.ts:129:24 - error TS2790: The operand of a 'delete' operator must be optional.
129                 delete (child as {parent: SParentElement}).parent;
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/base/model/smodel.ts:187:14 - error TS2610: 'index' is defined as an accessor in class 'SParentElement', but is overridden here in 'SModelRoot' as an instance property.
187     readonly index: SModelIndex<SModelElement>;
                 ~~~~~
src/features/edge-layout/model.ts:33:20 - error TS2339: Property 'hasFeature' does not exist on type 'never'.
33         && element.hasFeature(edgeLayoutFeature);
                      ~~~~~~~~~~
src/lib/model.ts:29:9 - error TS2611: 'anchorKind' is defined as a property in class 'SNode', but is overridden here in 'CircularNode' as an accessor.
29     get anchorKind() {
           ~~~~~~~~~~
src/lib/model.ts:38:9 - error TS2611: 'anchorKind' is defined as a property in class 'SNode', but is overridden here in 'RectangularNode' as an accessor.
38     get anchorKind() {
           ~~~~~~~~~~
[...]

PR #219, opened 2021-05-19
SPROTTY-77 fixed typscript 4 issues

BjBe82:
Fixed issues in Sprotty related to typescript 4

see: https://github.com/eclipse/sprotty/issues/218

### DIFF
--- a/src/base/model/smodel.ts
+++ b/src/base/model/smodel.ts
@@ -61,7 +61,7 @@ export class SModelElement {
     }
 
     get index(): SModelIndex<SModelElement> {
-        return this.root.modelRootindex;
+        return this.root.index;
     }
 
     /**
@@ -181,14 +181,17 @@ export class SChildElement extends SParentElement {
  * Base class for the root element of the diagram model tree.
  */
 export class SModelRoot extends SParentElement {
-    readonly modelRootindex: SModelIndex<SModelElement>;
     revision?: number;
 
     canvasBounds: Bounds = EMPTY_BOUNDS;
 
     constructor(index = new SModelIndex<SModelElement>()) {
         super();
-        this.modelRootindex = index;
+          // Override the index property from SModelElement, which has a getter, with a data property
+          Object.defineProperty(this, 'index', {
+            value: index,
+            writable: false
+        });
     }
 }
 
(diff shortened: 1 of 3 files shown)


BjBe82:
Causes severel errors

src/base/model/smodel.ts
```diff
@@ -110,7 +110,6 @@ export class SParentElement extends SModelElement {
             throw new Error(`No such child ${child.id}`);
         }
         children.splice(i, 1);
-        delete (child as {parent: SParentElement}).parent;
```
spoenemann, line 113: Instead of just omitting these lines, the `parent` property should at least be set to `undefined`.

src/features/edge-layout/model.ts
```diff
@@ -28,11 +28,15 @@ export interface EdgeLayoutable extends SModelExtension {
 export function isEdgeLayoutable<T extends SModelElement>(element: T): element is T & SChildElement & BoundsAware & EdgeLayoutable {
     return element instanceof SChildElement
         && element.parent instanceof SRoutableElement
-        && 'edgePlacement' in element
+        && checkEdgeLayoutable(element)
         && isBoundsAware(element)
         && element.hasFeature(edgeLayoutFeature);
 }
 
+function checkEdgeLayoutable(element:SChildElement): element is SChildElement& EdgeLayoutable{
```
spoenemann, line 36: The linter creates warnings because of inconsistent whitespace. Please run `yarn lint` and correct the warnings.

spoenemann (commented):
Thanks for the contribution! A few comments below.

src/base/model/smodel.ts
BjBe82: Basically, this is more or less the same as "delete". Because "parent” is defined as: `readonly parent: SParentElement;` which cannot be undefined this is not allowed. As mentioned here: https://github.com/eclipse/sprotty/issues/218 it would be required to define it as `parent?: SParentElement;` which effects a lot of code that needs to add "undefined" checks.

src/features/edge-layout/model.ts
BjBe82: fixed inconsistent whitespace

spoenemann (approved):
Thanks!

PR #223, opened 2021-05-29
Migrate to eslint

tohosaku:
tslint will be deprecated in 2019, prompting a move to eslint(https://github.com/palantir/tslint/issues/4534).

In the meantime, we propose to introduce eslint in a way that minimizes the modification of existing code.


spoenemann:
Thank you for the contribution!

I get an error when I run `yarn lint`:
```
$ eslint -c configs/.eslintrc.js src/**/!(*.spec.ts*)
/bin/sh: 1: Syntax error: "(" unexpected
error Command failed with exit code 2.
```
Do you have an idea where that comes from?

tohosaku:
Sorry. I forgot the quotes in the script. (Because this still works in Windows Powershell).
I have confirmed that it also works with bash, etc.

spoenemann (approved):
Great!

Issue #208, opened 2021-02-24
Line jumps to visually clarify intersecting lines


planger:
We'd like to add support for "line jumps" or "bridges" in order to guide the users' eye in following intersecting edges, as e.g. below:

![image](https://user-images.githubusercontent.com/588090/109037136-ac3bbe80-76ca-11eb-92cf-7671d4ab312b.png)

I'm wondering what would be the best way of implementing them in Sprotty, as an implementation for such line jumps would require awareness of the routes of all edges in a diagram.

Typically, for each `SRoutableElement` a `RoutableView` is invoked one after the other, which in turn runs the `IEdgeRouter` to obtain routes. These routes are then translated into a SVG path by the view. The routes of other edges, however, are not explicitly persistent or available to others during a rendering pass of the entire diagram, neither for the views nor the router. There would have to be a rendering context that is renewed for each rendering pass and that collects the routes of all edges rendered so far so that we could integrate a detection for intersections and subsequently rendering jumps for them. Also, it'd be nice to have these line jumps independently from the actual routers, as those jumps are not necessarily tied to the behavior of a router.

So another way could be to add a `IVNodePostprocessor`, which could pass through edges and find intersections and replace them with those line jumps. This would be nice because it would be independent of the actual router or even views (to some extent), however it would have to run the registered routers again for the model elements to compute their routes, which is inefficient, and then it would have to fiddle around with the path on SVG level, which seems a bit off too.

@spoenemann What do you think would be the best approach? Thanks a lot for your thoughts on that!

planger:
Right, good idea!

We could annotate the `vnode` with the segments via `svgPath.data.attachData`, react in a post processor to the graph root element and then run through all the edge segments to find intersections and then manipulate their SVG path. That could work... I'll give it a try in the next few days.

Thanks!

spoenemann:
Another approach: the `renderChildren` method has an optional `args` parameter:

https://github.com/eclipse/sprotty/blob/4ff859071e2b32ed94e031a973246037e71992ee/src/base/views/viewer.tsx#L79

The parent view that contains connections could pre-compute all routes and pass them as `args` so the contained connections can reuse that data. Then the connection views could include the "jumps" themselves when rendering. This approach would require only a few utility functions (pre-compute routes, render hops).

planger:
That sounds even better, thanks!
It'll make it easier to render the hops in the views (instead of having to manipulate the SVG after the fact).

spoenemann:
[...]

PR #226, opened 2021-07-06
Line jumps to visually clarify intersecting lines

planger:
- Allow to centralize computation of routes
- Introduce `EdgeRoutePostprocessor` which can work on routes after they've been computed
- Introduce `IntersectionFinder` as an `EdgeRouteProcessor` which detects and annotates intersections
- Add specialized `PolylineEdgeView` that renders jumps on the more horizontal edge of intersections and short breaks on the less horizontal edges

line-jumps

https://github.com/eclipse/sprotty/issues/208


src/features/edge-intersection/intersection-finder.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (a.intersectionPoint.x === b.intersectionPoint.x) {
+        return a.intersectionPoint.y - b.intersectionPoint.y;
+    }
+    return a.intersectionPoint.x - b.intersectionPoint.x;
+};
+
+
+export class RouteSegment {
+    constructor(public routableId: string, public segmentIndex: number, public point: Point) { }
+}
```
spoenemann, line 49: General note: I'd propose to use plain objects for such data structures, described by an `interface` type, instead of a `class`. This applies to other data structures in Sprotty as well, e.g. actions. Maybe we should refactor those at some point.

Rationale: The concept of using classes for data structures comes from the influence the many years of Java development have left on us (at least I can tell that for me and @JanKoehnlein). But this is rather unnatural in the JavaScript world, where plain objects are predominant. I would still use classes for service implementations and structures with a lot of attached behavior like the internal SModel, though.

src/features/edge-intersection/sweepline.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+// Based on the sweepline implementation at https://github.com/rowanwins/sweepline-intersections
+// which is published under the terms of MIT, but has been adapted to the use case of sprotty.
+import TinyQueue from "tinyqueue";
+import { Point, PointToPointLine } from "../../utils/geometry";
+import { Intersection } from "./intersection-finder";
+import { RoutedPoint } from "../routing/routing";
+
```
spoenemann, line 22: Please add comments to the algorithms in this file explaining the basic idea and with links to further information.

I know there's a lot of poorly commented code in this repository, but we need to start improving at some point :-)

src/graph/views.tsx
```diff
@@ -140,10 +235,10 @@ export class SLabelView extends ShapeView {
 
 @injectable()
 export class SCompartmentView implements IView {
-    render(compartment: Readonly<SCompartment>, context: RenderingContext): VNode | undefined {
+    render(compartment: Readonly<SCompartment>, context: RenderingContext, args?: object): VNode | undefined {
         const translate = `translate(${compartment.bounds.x}, ${compartment.bounds.y})`;
         const vnode = <g transform={translate} class-sprotty-comp="{true}">
-            {context.renderChildren(compartment)}
+            {context.renderChildren(compartment, args)}
```
spoenemann, line 241: The necessity to explicitly pass the `args` through all views is a change that needs to be explained, at least in the CHANGELOG.

I'm thinking whether we should make more explicit how the parameter is used. Maybe make the `args` parameter mandatory? Or define a more concrete type than `object`?

spoenemann (commented):
Thanks @planger! I came back from my vacation and finally found time to look into this. Generally it looks good, a few remarks below.

src/graph/views.tsx
planger: Yeah, good point! We should definitely mention that. I'll add this to the readme.

Whether to really enforce passing arguments or not: I think I'd be in favor to make the recommendation of passing those arguments on to the rendering of its children more explicit. However, as it'll break practically all existing `IView` implementations I'm a bit hesitant whether it is worth it and whether a recommendation in the Readme wouldn't be enough. What do you think?

src/features/edge-intersection/sweepline.ts
planger: Sure! I'll add some hopefully helpful doc.

src/features/edge-intersection/intersection-finder.ts
planger: Yeah, you're totally right. I tend to fall back to classes inadvertently.
Actually, however, I realized that `RouteSegment` is actually never used, so this is a left-over. Sorry about that.

tortmayr: :+1: I have created a separate issue (https://github.com/eclipse/sprotty/issues/233) to track the potential refactoring of action classes to `interfaces`:

src/graph/views.tsx
spoenemann: What do you think about this idea:
 * Add the following property to `RenderingContext` and to the constructor of `ModelRenderer`:
```typescript
readonly parentArgs?: object;
```
 * Change `renderChildren` to this:
```typescript
const context = args === undefined ? this : new ModelRenderer(this.viewRegistry, this.targetKind, this.postprocessors, args);
return element.children
    .map(child => context.renderElement(child, args))
    .filter(vnode => vnode !== undefined) as VNode[];
```
 * Change `renderElement` to this:
```typescript
const view = this.viewRegistry.get(element.type);
const vnode = view.render(element, this, this.parentArgs);
```
We could also consider merging the new `args` with the previous `parentArgs` to support properties reaching through to nested elements even when another view adds more properties.

And to make the whole thing more type-safe and be more consistent with React, we could renams `args` to `props` and change `IView` to this:
```typescript
export interface IView<P = {}> {
    render(model: Readonly<SModelElement>, context: RenderingContext, P: object): VNode | undefined
}
```
AFAICT these changes would be backwards-compatible.

planger: On a first glance this feels like a very good idea! Thanks!

Another idea instead of directly merging the `args` could be to always add them as `parentArgs` (args = `{...args, parentArgs: this.parentArgs}`)? This would most likely avoid unexpected property name clashes and consumers of the args could recursively query specific properties (such as the pre-computed routes which are on the root `args`)? Sprotty could provide a util function for that, similar to `findParentByFeature`.

Regarding the `props`... I like the idea of the generic type making the args more typesafe. Regarding the rename to `props`, I don't have a hard opinion, but I feel like the `model` corresponds to react's concept of `props` a bit more than `args`.

From your suggestions I took that you actually already have those changes locally, right? Do you want to add a commit to this PR or should I work on that?

spoenemann: > Another idea instead of directly merging the args could be to always add them as parentArgs (args = {...args, parentArgs: this.parentArgs})? This would most likely avoid unexpected property name clashes and consumers of the args could recursively query specific properties (such as the pre-computed routes which are on the root args)? Sprotty could provide a util function for that, similar to findParentByFeature.

Ok that seems like the most robust solution, but it needs a utility function to get the root arguments. Let's go for it.

> Regarding the props... I like the idea of the generic type making the args more typesafe. Regarding the rename to props, I don't have a hard opinion, but I feel like the model corresponds to react's concept of props a bit more than args.

Ok let's keep `args` for now.

> From your suggestions I took that you actually already have those changes locally, right? Do you want to add a commit to this PR or should I work on that?

No I don't have any local changes, the proposals were more like a scratchpad.

planger:
@spoenemann Thanks again for your feedback! I pushed a few more changes to address your comments, I believe.
Also I've added an alternative implementation of a view that just renders gap instead of jumps.

Peek 2021-09-03 21-30

Summary of changes from the commit comment:
```
This allows omitting to pass on the specified args when `renderChildren`
is invoked. The args are always passed on via the `RenderingContext`.
If new args are specified when invoking `renderChildren`, the parent
args are still available via `args.parentArgs`. To simplify finding
a value in the args or any of its parent args, we introduce the function
`findArgValue`.

This change also introduces a type for those args `IViewArgs`
which is essentially any object, but indicates that it may have
`parentArgs`. While views may still use `object` the usage of
`IViewArgs` makes it more clear what to expect. Views may
also choose to use a more specific type.
```

Please let me know what you think! Thanks!

src/base/views/viewer.tsx
```diff
@@ -59,19 +60,21 @@ export class ModelRenderer implements RenderingContext {
...
             return this.decorate(vnode, element);
         } else {
             return undefined;
         }
     }
 
-    renderChildren(element: Readonly<SParentElement>, args?: object): VNode[] {
+    renderChildren(element: Readonly<SParentElement>, args?: IViewArgs): VNode[] {
+        const childArgs = args ? { ...args, parentArgs: this.args } : this.args;
+        const context = new ModelRenderer(this.viewRegistry, this.targetKind, this.postprocessors, childArgs);
```
spoenemann, line 75: To avoid creating unnecessary new instances of ModelRenderer, we should reuse `this` if `args` is undefined.

src/base/views/viewer.tsx
```diff
@@ -59,19 +60,21 @@ export class ModelRenderer implements RenderingContext {
...
         return element.children
-            .map(child => this.renderElement(child, args))
+            .map(child => context.renderElement(child, childArgs))
```
spoenemann, line 77: Using `childArgs` both as function argument and ModelRenderer property seems redundant and potentially confusing to me. Shall we skip the `args` argument in `renderElement` and just use `this.args` in that method?

src/base/views/viewer.tsx
```diff
@@ -211,9 +218,9 @@ export class HiddenModelViewer implements IViewer {
     @inject(TYPES.ILogger) protected logger: ILogger;
 
     constructor(@inject(TYPES.ModelRendererFactory) modelRendererFactory: ModelRendererFactory,
-                @inject(TYPES.PatcherProvider) patcherProvider: PatcherProvider,
-                @multiInject(TYPES.HiddenVNodePostprocessor) @optional() hiddenPostprocessors: IVNodePostprocessor[]) {
-        this.hiddenRenderer = modelRendererFactory('hidden', hiddenPostprocessors);
+        @inject(TYPES.PatcherProvider) patcherProvider: PatcherProvider,
+        @multiInject(TYPES.HiddenVNodePostprocessor) @optional() hiddenPostprocessors: IVNodePostprocessor[]) {
+        this.hiddenRenderer = modelRendererFactory('hidden', hiddenPostprocessors, { root: true, hidden: true });
```
spoenemann, line 223: Do you use the `root` and `hidden` properties anywhere?
`hidden` seems redundant to `targetKind`.

src/lib/svg-views.tsx
```diff
@@ -31,19 +31,19 @@ import { injectable } from 'inversify';
 
 @injectable()
 export class SvgViewportView implements IView {
-    render(model: Readonly<ViewportRootElement>, context: RenderingContext): VNode {
+    render(model: Readonly<ViewportRootElement>, context: RenderingContext, args?: IViewArgs): VNode {
         const transform = `scale(${model.zoom}) translate(${-model.scroll.x},${-model.scroll.y})`;
         return <svg>
             <g transform={transform}>
-                {context.renderChildren(model)}
+                {context.renderChildren(model, args)}
```
spoenemann, line 38: Passing the `args` explicitly is no longer necessary, right?
See also other views below.

src/base/views/viewer.tsx
planger: Right, I appended a commit that avoids creating new objects if `args` is `undefined`.

src/base/views/viewer.tsx
planger: Agreed, I removed the `childArgs` argument from the `renderElement` call.

src/base/views/viewer.tsx
planger: No, I don't use them. I just thought they may be good to have, but this was rather speculative and you are right, that this is actually redundant to `targetKind`.

src/lib/svg-views.tsx
planger: Right, they are not needed anymore. I must have missed those, as I believe I removed them anywhere else. In the new commit I removed them from the `svg-views.tsx` too. Thanks!

src/base/views/viewer.tsx
spoenemann: Shall we also remove the `args` from the `renderElement` signature then?

src/base/views/viewer.tsx
spoenemann: Isn't `args.root === true` equivalent to `args. parentArgs === undefined`? In that case we could remove `root` as well.

planger: Right, I just put it in there to avoid having an empty root args object. But it doesn't add real value per se. So you're right, we can remove it.

spoenemann (approved):
Thanks!

PR #228, opened 2021-07-14
Enabled eslint rule "no-prototype-builtins".

tohosaku:
Eslint warns calling the hasOwnProperty method directly from object.
Where we can do without hasOwnProperty, I remove it, and use the own functions if needed.

Also, added type guard so that "as any" does not have to be used.

### DIFF
--- a/src/utils/object.ts
+++ b/src/utils/object.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2019 TypeFox and others.
+ * Copyright (c) 2017-2021 TypeFox and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -14,6 +14,14 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-export function isInjectable(constr: new (...args: unknown[]) => unknown): boolean {
-    return Reflect.getMetadata('inversify:paramtypes', constr) !== undefined;
+ export function isNotNullish(data: unknown): data is Record<PropertyKey, unknown> {
+    return data !== null && data !== undefined;
+}
+
+export function hasOwnProperty<K extends PropertyKey>(arg: unknown, key: K): arg is Record<K, unknown> {
+    return isNotNullish(arg) && Object.prototype.hasOwnProperty.call(arg, key);
+}
+
+export function safeAssign<T>(target: T, partial: Partial<T>): T {
+    return Object.assign(target, partial);
 }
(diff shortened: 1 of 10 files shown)


src/utils/object.ts
```diff
@@ -0,0 +1,27 @@
+/********************************************************************************
+ * Copyright (c) 2017-2021 TypeFox and others.
```
spoenemann, line 2: Please use the current year for new files, and change the affiliation to your own if you like.

src/utils/object.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+ export function isNotNullish(data: unknown): data is Record<PropertyKey, unknown> {
+    return data !== null && data !== undefined;
```
spoenemann, line 18: A string value meets this check, but is not a `Record<PropertyKey, unknown>`.

Shouldn't this rather do
```typescript
return typeof data === 'object' && data !== null;
```
Maybe the function should be renamed to `isObject` then.

src/utils/object.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export function hasOwnProperty<K extends PropertyKey>(arg: unknown, key: K): arg is Record<K, unknown> {
```
spoenemann, line 21: Many usages of this function include checking the type of the property. Would it be cool to add an optional third parameter `type`? Then you could even change the `unknown` in the return type to something more specific. Here's a proposal:

```typescript
export type TypeOf<T> =
    T extends number ? 'number'
    : T extends string ? 'string'
    : T extends boolean ? 'boolean'
    : T extends bigint ? 'bigint'
    : T extends symbol ? 'symbol'
    : T extends Function ? 'function'
    : T extends object ? 'object'
    : 'undefined';

export function hasOwnProperty<K extends PropertyKey, T>(arg: unknown, key: K, type?: TypeOf<T> | ((v: unknown) => v is T)): arg is Record<K, T> {
    if (!(isObject(arg) && Object.prototype.hasOwnProperty.call(arg, key))) {
        return false;
    }
    if (typeof type === 'string') {
        return typeof arg[key] === type;
    }
    if (typeof type === 'function') {
        return type(arg[key]);
    }
    return true;
}
```

spoenemann:
@tohosaku your contributions are very valuable for Sprotty's code quality! Would you like to become a committer of the project? That would give you push access (so you don't have to use a fork to create PRs) and list you on this page:

https://projects.eclipse.org/projects/ecd.sprotty/who

tohosaku:
Sorry for the delay in replying. I'm honored to be invited to be a committer. Please let me join.

src/utils/object.ts
tohosaku: I've incorporated it. The caller's code becomes clearer.

spoenemann (approved):

PR #234, opened 2021-08-12
Add dependency-injection support to LayoutRegistry

CamilleLetavernier:
Modify the LayoutRegistry to support LayoutRegistrations via dependency-injection, similar to how it's done for other registries

### DIFF
--- a/src/features/bounds/layout.ts
+++ b/src/features/bounds/layout.ts
@@ -14,7 +14,7 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import { inject, injectable, multiInject, optional } from "inversify";
+import { inject, injectable, interfaces, multiInject, optional } from "inversify";
 import { TYPES } from "../../base/types";
 import { ILogger } from '../../utils/logging';
 import { InstanceRegistry } from "../../utils/registry";
@@ -22,6 +22,7 @@ import { Bounds, EMPTY_BOUNDS } from "../../utils/geometry";
 import { SParentElement, SModelElement } from "../../base/model/smodel";
 import { isLayoutContainer, LayoutContainer } from "./model";
 import { BoundsData } from "./hidden-bounds-updater";
+import { isInjectable } from "../../utils/inversify";
 
 @injectable()
 export class LayoutRegistry extends InstanceRegistry<ILayout> {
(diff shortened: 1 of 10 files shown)


CamilleLetavernier:
I'm not sure how the default layouts should be registered. Ideally, we want to be able to remove them, or to associate a different implementation with the same key (e.g. providing our own implementation for the "vbox" layout).

src/features/bounds/layout.ts
```diff
@@ -14,27 +14,34 @@
...
 export class LayoutRegistry extends InstanceRegistry<ILayout> {
-    constructor() {
+    constructor(@multiInject(TYPES.LayoutRegistration) @optional() layouts: (LayoutRegistration)[] = []) {
         super();
-        this.register(VBoxLayouter.KIND, new VBoxLayouter());
-        this.register(HBoxLayouter.KIND, new HBoxLayouter());
-        this.register(StackLayouter.KIND, new StackLayouter());
+        layouts.forEach(layout => {
+            if (this.hasKey(layout.layoutKind)) {
+                console.log('Layout kind is already defined: ', layout.layoutKind);
```
spoenemann, line 32: Log messages should be printed via the `ILogger` service.

spoenemann (commented):
Hi @CamilleLetavernier, thanks for the contribution!

You can look at the action handler registrations to see how to define the registration in a way that the default layouts can be overridden:

spoenemann:
Yes, that makes sense! In that case it can be expressed more simply:
```typescript
bind(MoveMouseListener).toSelf().inSingletonScope();
bind(TYPES.MouseListener).toService(MoveMouseListener); 
```

CamilleLetavernier:
Thanks for the reviews!

I've addressed the simple comments (Using logger & sorting the TYPES). I also noticed that the tests were broken, because manually instantiating a LayoutRegistry no longer registers the 3 default layouts; it should be fixed now.

I still need to add a "configureLayout" function, similar to the Action handlers, but I won't have time to do that (and test that with my use cases) properly this week; so I'll keep it for later :)

spoenemann (approved):
Great!

spoenemann:
> Maybe we should apply this pattern to all classes that are multiinjected. I haven't had a detailed look yet, but I know that there at least some multiinjected classes that are not bound to itself before they are bound to the multiinjected key.

@tortmayr is there already an issue about this?

tortmayr:
> > Maybe we should apply this pattern to all classes that are multiinjected. I haven't had a detailed look yet, but I know that there at least some multiinjected classes that are not bound to itself before they are bound to the multiinjected key.
> 
> @tortmayr is there already an issue about this?

I don't think so. I created https://github.com/eclipse/sprotty/issues/239 to track this.

Issue #25, opened 2018-11-23
[performance] take a shortcut for viewport changes


JanKoehnlein:
With big diagrams, scrolling and zooming gets very slow as we recalculate and patch the entire tree.
We should add so kind of shortcut for changes that just affect the viewport.

FelixJoehnk:
Hello,
I am currently working on icreasing the  performace. The Main Problem for large graphs are Zoom actions right now, resulting in  setviewport actions  that are done one after another.
The Idea we had was to make a delay (5-20ms) so that one can make a list of setviewport actions in the Mouse Tool and then merge those in a good way :)
Is that a change that should bei added to sprotty?
Then i would try and rewrite Part of the Mouse Tool !

Greate regards Felix

spoenemann:
The `merge` method is used to merge multiple commands within the same animation frame:
https://github.com/eclipse-sprotty/sprotty/blob/6d43db24c67c4adc837ea9e8a775f47ded5c5193/packages/sprotty/src/base/commands/command.ts#L108-L118

As to the proposal to take a shortcut for viewport changes: This would break the implementation of Level-of-Detail rendering, i.e. omitting selected details when the user has zoomed out. We already have that in our Multicore example.

There are already other ways to improve the performance for large graphs, and those work for all diagram interactions, not just viewport changes. I had a presentation about that together with @jbicker at EclipseCon 2023, and we're going to publish a blog post at https://www.typefox.io/blog/ soon.

PR #237, opened 2021-08-19
Zoomactions Merged & Typescript 4.3

FelixJoehnk:
Hy i have made some addition in the MouseTool aswell as the Zoom MouseListener. The Idea was to execute one zoom action then wait 5ms afterwards with the next zoom action so pend up wheel events may be stored in a array. Then the entire array will be used as one zoom action resulting in an overall improved performance (see #25). Also i changed the typescript version to 4.3 and changed all necessary things (hopefully).

### DIFF
--- a/src/base/views/mouse-tool.ts
+++ b/src/base/views/mouse-tool.ts
@@ -33,6 +33,11 @@ export class MouseTool implements IVNodePostprocessor {
 
     constructor(@multiInject(TYPES.MouseListener) @optional() protected mouseListeners: MouseListener[] = []) { }
 
+    delay = (ms: number) => new Promise(res => { setTimeout(res, ms); });
+    delayed = false;
+    delayedWEvents: WheelEvent[] = [];
+    delayedWModels: SModelRoot[] = [];
+
     register(mouseListener: MouseListener) {
         this.mouseListeners.push(mouseListener);
     }
@@ -79,6 +84,32 @@ export class MouseTool implements IVNodePostprocessor {
         }
     }
 
+    protected handleEvents<K extends keyof MouseListener>(methodName: K, models: SModelRoot[], events: MouseEvent[]) {
+        let elements: (SModelElement | undefined)[] = [];
+        this.focusOnMouseEvent(methodName, models[0]);
+        for (let i = 0; i < events.length; i++) {
+            elements.push(this.getTargetElement(models[i], events[i]));
+        }
+
+        if (elements.length > 0) {
+            const actions = this.mouseListeners
+                .map(listener => listener[methodName].apply(listener, [elements, events]))
+                .reduce((a, b) => a.concat(b));
+            if (actions.length > 0) {
+                events[0].preventDefault();
+                for (const actionOrPromise of actions) {
+                    if (isAction(actionOrPromise)) {
+                        this.actionDispatcher.dispatch(actionOrPromise);
+                    } else {
+                        actionOrPromise.then((action: Action) => {
+                            this.actionDispatcher.dispatch(action);
+                        });
+                    }
+                }
+            }
+        }
+    }
+
     protected focusOnMouseEvent<K extends keyof MouseListener>(methodName: K, model: SModelRoot) {
         if (document) {
             const domElement = document.getElementById(this.domHelper.createUniqueDOMElementId(model));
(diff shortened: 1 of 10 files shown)


src/base/views/mouse-tool.ts
```diff
@@ -79,6 +84,32 @@ export class MouseTool implements IVNodePostprocessor {
         }
     }
 
+    protected handleEvents<K extends keyof MouseListener>(methodName: K, models: SModelRoot[], events: MouseEvent[]) {
```
NiklasRentzCAU, line 87: To prevent code duplication, the handleEvent function should now call this handleEvents function with single-element arrays.

src/features/viewport/zoom.ts
```diff
@@ -61,6 +61,37 @@ export class ZoomMouseListener extends MouseListener {
         return [];
     }
 
+    wheels(targets: SModelElement[], events: WheelEvent[]): Action[] {
```
NiklasRentzCAU, line 64: Same here, the wheel function may as well just call the wheels function now with single-element lists.

src/base/views/mouse-tool.ts
```diff
@@ -190,6 +238,10 @@ export class MouseListener {
         return [];
     }
 
+    wheels(target: SModelElement[], events: WheelEvent[]): (Action | Promise<Action>)[] {
```
NiklasRentzCAU, line 241: To avoid breaking the code of users not having implemented the wheels function yet, this should have a default implementation repeatedly calling the wheel function, collecting and returning the actions.

NiklasRentzCAU (changes requested):
Looks to be a great addition to tackle zooming and panning in 'slow' diagrams to queue up and execute one after the other but instead to merge together and be more responsive. If you would have a look at the comments I think we should include this in the next release.

@spoenemann Would you also have a look at this please? The issue tackled here was obvious in some of our larger diagrams and needed a fix, this seems to be the right place for it.

src/base/views/mouse-tool.ts
FelixJoehnk: I made the change for now but this is not as easy as it seems...
The wheel function might use the ```viewport = findParentByFeature(targets[i], isViewport);``` to access the current viewport but since the viewport wont change between the executed wheel actions (when exec from wheels) it will never change and might therefor break users code nonetheless

src/base/views/mouse-tool.ts
FelixJoehnk: This seems a good change.

src/features/viewport/zoom.ts
FelixJoehnk: also seems to be good

spoenemann:
I'm not sure the event listener is the best place to combine actions. We already have the viewer cache, which combines model changes and applies them once per animation frame:
https://github.com/eclipse/sprotty/blob/master/src/base/views/viewer-cache.ts

In my experience the main bottleneck is not the processing made by Sprotty, but the actual DOM rendering by the browser. So the performance of large diagrams can be greatly boosted by removing invisible or hardly visible elements from the DOM. I prepared the groundwork for that a year ago (#182). It enables skipping elements that are outside the viewport. In addition, it is easy to add conditional rendering to small elements that appear frequently so they are skipped or replaced by a simpler shape when zoomed out. For example, you can skip all labels when the zoom factor exceeds a certain threshold, or replace them with gray rectangles.

FelixJoehnk:
So i have changed the location. If i understand the ViewerCache right it will normally update on every rendering frame or if there is any action which requires changes(cause !== undefined). It Should therefor be enough to do the rendering only on every rendering frame for zoom / panning actions  (cause !== SetViewportAction).
 Another thing i would like to mention is that even if one removes the DOM elements that are out of bounds or which are simply to small the idea to merge zoom actions may still be needed if one has a flood of ZoomActions(SetViewPortActions). Or the time needed to render the visible DOM elements is too large since many might fit in a small area and still be visible enough to get rendered.

FelixJoehnk:
Sorry i have no idea what went wrong there with that merge ...
However now it should work!
@NiklasRentzCAU  I think you are right! One might create such things as commands who alter the model without any actions i am unsure if the moving of single nodes in sprotty has a corresponding action. 
As far as i understand: the viewer-cache schedules a update for the moment where the browser requests a new frame. The Model which should then be displayed gets updated everytime a new model arrives which has and no cause for the update given. If in the meantime the model got updated with any cause the scheduled update gets a undefined model resulting in no update. 
Here i think one can simply say that if it is a setviewportaction one should wait for the next animationframe wanted by the browser to do it and if a new model arrives just use that model for the update instead.
However this results in a small time window where one waits for the next frame where a browser requests a new animationFrame but for me its not that noticable even on small graphs. For larger graphs this delay is not relevant since one waits longer for the actual rendering of the DOM then the browser to call for a new animationframe.
Also if merging SetViewportActions this way one only processes the last action(for example in the ModelViewer). if users altered these to do something for these actions it might break some things for them.

spoenemann:
Are you sure the new condition is ever met? AFAICT the `cause` is set only when the command returns a `CommandResult`  object where the corresponding property is explicitly set. Of the default commands, only `RequestBoundsCommand` and `ExportSvgCommand` do this.

There is an additional mechanism to merge multiple modifications into one: `SetViewportCommand` implements the `merge` method, so subsequent instances should automatically be reduced to one.

FelixJoehnk:
You are right however i noticed that Changes in the current master Branche fixed the issuei had
Seems that when using a Touchpad and zooming in, the Model is moved too and therefore the Zooms are not mergeable anymore... 
Like i Said hawever this is already fixed in the master Branche. maybe you could publish a new Version of the master :)

Great regards Felix

PR #240, opened 2021-09-08
Implemented projection on horizontal and vertical bars

spoenemann:
This PR adds "projection bars". These are similar to the scrollbars in IDEs like Eclipse and VS Code, where you have colored markers at interesting code regions like errors or SCM changes. The difference is that they can appear in two dimensions, so you can have a horizontal and a vertical projection.

Projection bars are added by using `ProjectedViewportView` as view of the root element and configuring marker colors for some model elements. This can be either dynamic by implementing `getProjection` (for example to show only elements with certain properties) or static with the following pattern in the diagram configuration:
```typescript
configureModelElement(context, 'pre-logo', ShapedPreRenderedElement, PreRenderedView, {
    viewInit: setProjection('rgba(255, 153, 0, 0.5)')
});
```
An additional requirement of the projection bars is that the root element needs to be BoundsAware. its position and size are used to scale the content of the bars properly.

I applied this new feature to the "SVG" example. Try zooming in and clicking / dragging the scroll bars, and see how they are updated when you move elements around.

### DIFF
--- a/examples/svg/src/di.config.ts
+++ b/examples/svg/src/di.config.ts
@@ -20,8 +20,7 @@ import {
     ProjectedViewportView, ViewportRootElement, ShapedPreRenderedElement, configureModelElement,
     ForeignObjectElement, ForeignObjectView, RectangularNode, RectangularNodeView, moveFeature,
     selectFeature, EditableLabel, editLabelFeature, WithEditableLabel, withEditLabelFeature,
-    isEditableLabel, Action, MoveAction, SShapeElementSchema, ViewportRootElementSchema,
-    ActionHandlerRegistry, SetBoundsAction
+    isEditableLabel
 } from '../../../src';
 
 export default () => {
@@ -32,7 +31,7 @@ export default () => {
     const svgModule = new ContainerModule((bind, unbind, isBound, rebind) => {
         rebind(TYPES.ILogger).to(ConsoleLogger).inSingletonScope();
         rebind(TYPES.LogLevel).toConstantValue(LogLevel.log);
-        bind(TYPES.ModelSource).to(SVGModelSource).inSingletonScope();
+        bind(TYPES.ModelSource).to(LocalModelSource).inSingletonScope();
         const context = { bind, unbind, isBound, rebind };
         configureModelElement(context, 'svg', ViewportRootElement, ProjectedViewportView);
         configureModelElement(context, 'pre-rendered', ShapedPreRenderedElement, PreRenderedView);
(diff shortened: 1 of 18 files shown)


examples/svg/src/di.config.ts
```diff
@@ -48,6 +57,61 @@ export default () => {
     return container;
 };
 
+export class SVGModelSource extends LocalModelSource {
```
spoenemann, line 60: If this code turns out to be useful for others, we could move it to the sprotty package.

planger:
Awesome, thanks for the nice work @spoenemann! This is an excellent new feature. I haven't worked through the code in detail yet, but plan to do so in the next days.

However, I wanted to get your feedback on two thoughts already.

1. I was wondering whether an element/view type that shall show a projection wouldn't be a good fit for an `SModelExtension` introducing the `projectionColor` and enforcing to be a `BoundsAware` too. It would make the new property in `IView` implementations superfluous and would probably avoid introducing another extension concept for views (`viewInit`) -- even though this view initialization concept might be handy anyway. Obtaining the projections in the `ModelRenderer` would even be easier as it wouldn't have to obtain the view registry just to get the elements' projections.
2. It might be useful to let these "projectable" views (or model elements, see 1.) rather specify CSS classes instead or in addition to the `projectionColor` style string. This could yield more flexibility (borders, corner styles, even anomations, etc.) and would likely also make it easier to integrate with themes and react to theme changes in e.g. Theia or VSCode.

Just my first two thoughts, I'll work through the code in the next days.

Thanks again!

spoenemann:
Good thoughts, I'll look into that and notify you again.

spoenemann:
@planger I changed the new code as you proposed, it's actually much cleaner now.

planger (approved):
Code looks great now, thanks @spoenemann! This is an awesome feature!

In more detailed testing I found it a bit confusing though that the projections don't scale down when zooming out beyond the viewport size.
If zooming in, the blue rectangles indicating the current visible excerpt of the diagram make it somewhat clear how the visible viewport relates to the projections. But when zooming out, the blue rectangles remain filling up 100% obviously (where else should they go ;)), but then the projections become difficult to understand. Wouldn't it become more clear, if the blue rectangles together with the projections would scale down once you zoom out beyond viewport size with just white space left/right/top/bottom filling the void in the toolbars that represents the visible area beyond the viewport? 

Also I could imagine that it would be nice, if double-clicking a projection would center the projected element. This would play nicely with the single-click, which sets one dimension of the projection, double-clicking could set both dimensions.

planger (approved):
Excellent, thanks a lot for the update!
I think scaling the projections really help a lot in comprehending the projections if zoomed out a lot. Also thanks for extracting the `getModelBounds()` method.

Code looks great too!

Just for the reference, a gif for others to see how it looks like:
projections

spoenemann:
I added the double-click behavior. @planger would you like to test that, too?

planger:
> I added the double-click behavior. @planger would you like to test that, too?

Sure, works like a charm! Thanks for adding this nice feature!

PR #249, opened 2021-10-14
#248 Support VSCode Codicons

### LINKED ISSUE #248: Support VSCode Codicons
- Support Cocidons via [`@vscode/codicons`](https://github.com/microsoft/vscode-codicons)
- Provide util methods to create CSS classes
- Keep FontAwesome icons and show that both icon sets can be used in example

ndoschek:
Fixes #248 

- Support Codicons via @vscode/codicons
  - Migrate command-palette to codicons
  - Fix loading indicator in command-palette
- Provide util methods to create codicon CSS classes
- Keep FontAwesome icons and show that both icon sets can be used in example
  - Class diagram example: command palette uses codicons, class popup uses font-awesome icon (info-circle)


ndoschek:
Ah you are right, thanks for catching that! I opened a PR for that: https://github.com/eclipse/sprotty/pull/250

PR #250, opened 2021-10-21
#248 Fix Codicons dependency

ndoschek:
Move `"@vscode/codicons"` from `devDependencies` to `dependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "url": "https://github.com/eclipse/sprotty.git"
   },
   "dependencies": {
+    "@vscode/codicons": "^0.0.25",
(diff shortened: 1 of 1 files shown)


spoenemann:
I'm thinking about whether we need to file a CQ for this dependency because it's [CC-licensed](https://github.com/microsoft/vscode-codicons/blob/main/LICENSE). But since [Theia is using it as well](https://github.com/eclipse-theia/theia/blob/38e2ba301785ec4a08dfc27a8b23d159f489bbdf/packages/core/package.json#L35), it's probably not necessary. WDYT?

PR #253, opened 2021-11-10
Created new package sprotty-protocol

### LINKED ISSUE #233: Consider refactoring actions to interface`s instead of classes
see also https://github.com/eclipse/sprotty/pull/226#discussion_r684017707
The concept of using classes for data structures comes from the influence the many years of Java development have left on us (at least I can tell that for me and @JanKoehnlein). But this is rather unnatural in the JavaScript world, where plain objects are predominant.

For actions using classes also has an additional downside. It tricks the user into using `instance of` for type checks which does not work for all actions. Namely, actions that are received from a lsp/glsp server or the sprotty-vscode-extensiosn are plain objects and an `instance of` check will fail even if the action is theoretically an instance of the class.

Even if we don't move away from the class-based approach we should at least ensure that we have a proper guard function for each action (`isXYZAction()`) and use them over instance of checks.

### LINKED ISSUE #243: Node.js support and restructuring to monorepo
I am seeing increasing demand for a Node.js backend for Sprotty. I already created an example for that about a year ago:

https://github.com/spoenemann/theia-sprotty-nodejs-example/blob/master/diagram-view/src/node/diagram-server-impl.ts

I would like to share this to the Sprotty project as a separate package. At the same time, I propose to change this repository to a monorepo with multiple packages:

 * `sprotty`
 * `sprotty-elk` (moved from [eclipse/sprotty-layout](https://github.com/eclipse/sprotty-layout))
 * `sprotty-node` (new)
 * `sprotty-protocol` (new)

This will make it considerably easier to keep the code synchronized and to implement cross-package functionality.

I'd leave `sprotty-theia` and `sprotty-vscode` separate because their dependency tree is more complex, and I have previously seen problems when trying to include both the Theia and the VS Code integration in the same Yarn workspace. I would also keep `sprotty-server` separate to avoid a mixed-platform repository (TypeScript / Java).

`sprotty-node` should not have a dependency to `sprotty` to avoid using browser code in a Node.js context. For that reason, we need to extract common data structures into a new package `sprotty-protocol`. This is also a good opportunity to isolate and document the client / server protocol of Sprotty more clearly. For now that new repository should contain

 * SModel (schema) declarations
 * Actions (#233)

In order to keep API breakage at a minimum, we could mark the original declarations in the `sprotty` package as deprecated (they will still work, but should be changed to the new `sprotty-protocol` definitions at the earliest convenience).

What do you think?

spoenemann:
Closes #233, closes #243. This is a large change, and I certainly don't expect anyone to review it line-by-line, but any testing with existing applications would be appreciated.

I created a new package `sprotty-protocol` and *copied* the following aspects from `sprotty`. Where possible, the original declarations have been kept and marked `@deprecated`.

 * Many action declarations (now `actions.ts`)
 * Model element schema (now `model.ts`)
 * Most of `utils/geometry.ts`
 * `utils/async.ts`
 * `utils/json.ts`
 * `utils/object.ts`

### Actions

As discussed in #233, I refactored the action classes into interfaces with accompanying namespaces. This supports the notion of actions being lightweight serializable data structures that must not have any attached behavior. This is a breaking change, but only in the way how actions are created: instead of calling a constructor, you call the declared `create` function.

I left those actions that are used only within the client in the `sprotty` package, and also those where I wasn't sure. Feel free to move (i.e. copy and mark `@deprecated`) more action definitions to `sprotty-protocol` if you need them in a Node.js based server. This is something we can still do later without breaking user code: we would remove deprecated code only when we increase the major version.

### Model Elements

I copied all model element schema interfaces to `sprotty-protocol`, but removed the `Schema` suffix (e.g. `SNode` instead of `SNodeSchema`) because

 * it's shorter,
 * it's more consistent with our existing server code in Java,
 * the name "schema" is a bit misleading,
 * we use only the external model in the diagram server, but not the internal model.

Now there's a name conflict between the external model (schema) defined in `sprotty-protocol` and the internal model defined in `sprotty`. Possible options:

 * Keep it as it is now; in case you need to import both the internal and external model in one file, you can resolve the conflict at the import declaration (this should rarely be necessary in user code).
 * Add a suffix to internal model elements, e.g. `SNodeImpl`.
 * Change the prefix of internal model elements, e.g. `InternalNode`.

What do you think?

### Model Index

I split the `SModelIndex` in two classes:

 * `SModelIndex` in `sprotty-protocol` for indexing external models (only on demand)
 * `ModelIndexImpl` in `sprotty` for indexing internal models (attached to the root element)

### Diagram Server

I added a `DiagramServer` class to `sprotty-protocol` for implementing Node.js based diagram servers. This is similar to the diagram server in our Java code base. I renamed the previous `DiagramServer` in `sprotty` to `DiagramServerProxy`.

### DIFF
--- a/packages/sprotty-protocol/src/actions.ts
+++ b/packages/sprotty-protocol/src/actions.ts
@@ -1,0 +1,500 @@
+/********************************************************************************
+ * Copyright (c) 2021 TypeFox and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { SModelRoot, SModelElement, Viewport } from './model';
+import { Bounds, Point, Dimension } from './utils/geometry';
+import { JsonAny, JsonPrimitive } from './utils/json';
+import { hasOwnProperty } from './utils/object';
+
+/**
+ * Wrapper for actions when transferring them between client and server.
+ * The `clientId` is used to identify the specific diagram instance in the client.
+ */
+export interface ActionMessage {
+    clientId: string
+    action: Action
+}
+
+export function isActionMessage(object: unknown): object is ActionMessage {
+    return hasOwnProperty(object, 'action');
+}
+
+/**
+ * An action describes a change to the model declaratively.
+ * It is a plain data structure, and as such transferable between server and client.
+ */
+export interface Action {
+    kind: string;
+}
+
+export function isAction(object?: unknown): object is Action {
+    return hasOwnProperty<string, string>(object, 'kind', 'string');
+}
+
+/**
+ * A request action is tied to the expectation of receiving a corresponding response action.
+ * The `requestId` property is used to match the received response with the original request.
+ */
+export interface RequestAction<Res extends ResponseAction> extends Action {
+    requestId: string
+}
+
+export function isRequestAction(object?: Action): object is RequestAction<ResponseAction> {
+    return hasOwnProperty<string, string>(object, 'requestId', 'string');
+}
+
+let nextRequestId = 1;
+/**
+ * Generate a unique `requestId` for a request action.
+ */
+export function generateRequestId(): string {
+    return (nextRequestId++).toString();
+}
+
+/**
+ * A response action is sent to respond to a request action. The `responseId` must match
+ * the `requestId` of the preceding request. In case the `responseId` is empty or undefined,
+ * the action is handled as standalone, i.e. it was fired without a preceding request.
+ */
+export interface ResponseAction extends Action {
+    responseId: string;
+}
+
+/**
+ * A response action is sent to respond to a request action. The `responseId` must match
+ * the `requestId` of the preceding request. In case the `responseId` is empty or undefined,
+ * the action is handled as standalone, i.e. it was fired without a preceding request.
+ */
+
+export function isResponseAction(object?: Action): object is ResponseAction {
+    return hasOwnProperty<string, string>(object, 'responseId', 'string');
+}
+
+/**
+ * A reject action is fired to indicate that a request must be rejected.
+ */
+export interface RejectAction extends ResponseAction {
+    kind: typeof RejectAction.KIND
+    message: string
+    detail?: JsonAny
+}
+export namespace RejectAction {
+    export const KIND = 'rejectRequest';
+
+    export function create(options: { message: string, detail?: JsonAny, requestId: string }): RejectAction {
+        return {
+            kind: KIND,
+            message: options.message,
+            detail: options.detail,
+            responseId: options.requestId
+        };
+    }
+}
+
+/**
+ * Sent from the client to the model source (e.g. a DiagramServer) in order to request a model. Usually this
+ * is the first message that is sent to the source, so it is also used to initiate the communication.
+ * The response is a SetModelAction or an UpdateModelAction.
+ */
+export interface RequestModelAction extends RequestAction<SetModelAction> {
+    kind: typeof RequestModelAction.KIND
+    options?: { [key: string]: string | number | boolean }
+}
+export namespace RequestModelAction {
+    export const KIND = 'requestModel';
+
+    export function create(options?: { [key: string]: JsonPrimitive }): RequestModelAction {
+        return {
+            kind: KIND,
+            options,
+            requestId: generateRequestId()
+        };
+    }
+}
+
+/**
+ * Sent from the model source to the client in order to set the model. If a model is already present, it is replaced.
+ */
+export interface SetModelAction extends ResponseAction {
+    kind: typeof SetModelAction.KIND
+    newRoot: SModelRoot
+}
+export namespace SetModelAction {
+    export const KIND = 'setModel';
+
+    export function create(newRoot: SModelRoot, requestId?: string): SetModelAction {
+        return {
+            kind: KIND,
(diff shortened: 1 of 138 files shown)


packages/sprotty-protocol/src/actions.ts
```diff
@@ -0,0 +1,500 @@
...
+ * The response is a SetModelAction or an UpdateModelAction.
+ */
+export interface RequestModelAction extends RequestAction<SetModelAction> {
+    kind: typeof RequestModelAction.KIND
+    options?: { [key: string]: string | number | boolean }
+}
+export namespace RequestModelAction {
+    export const KIND = 'requestModel';
+
+    export function create(options?: { [key: string]: JsonPrimitive }): RequestModelAction {
```
tortmayr, line 119: Same as above: `JsonMap` type could be used here

packages/sprotty-protocol/src/actions.ts
```diff
@@ -0,0 +1,500 @@
...
+ * Sent from the model source to the client in order to set the model. If a model is already present, it is replaced.
+ */
+export interface SetModelAction extends ResponseAction {
+    kind: typeof SetModelAction.KIND
+    newRoot: SModelRoot
+}
+export namespace SetModelAction {
+    export const KIND = 'setModel';
+
+    export function create(newRoot: SModelRoot, requestId?: string): SetModelAction {
```
tortmayr, line 138: The create function for response actions should be initialized with an empty request id (to enable standalone response actions)
```suggestion
    export function create(newRoot: SModelRoot, requestId=''): SetModelAction {
```

packages/sprotty-protocol/src/actions.ts
```diff
@@ -0,0 +1,500 @@
...
+ * This action can also be used to remove any existing popup by choosing EMPTY_ROOT as root element.
+ */
+export interface SetPopupModelAction extends ResponseAction {
+    kind: typeof SetPopupModelAction.KIND
+    newRoot: SModelRoot
+}
+export namespace SetPopupModelAction {
+    export const KIND = 'setPopupModel';
+
+    export function create(newRoot: SModelRoot, requestId?: string): SetPopupModelAction {
```
tortmayr, line 221:
```suggestion
    export function create(newRoot: SModelRoot, requestId=''): SetPopupModelAction {
```

packages/sprotty-protocol/src/actions.ts
```diff
@@ -0,0 +1,500 @@
...
+export interface ComputedBoundsAction extends ResponseAction {
+    kind: typeof ComputedBoundsAction.KIND
+    bounds: ElementAndBounds[]
+    revision?: number
+    alignments?: ElementAndAlignment[]
+}
+export namespace ComputedBoundsAction {
+    export const KIND = 'computedBounds';
+
+    export function create(bounds: ElementAndBounds[], options: { revision?: number, alignments?: ElementAndAlignment[], requestId: string }): ComputedBoundsAction {
```
tortmayr, line 287: RequestId should be an optional option:
```suggestion
    export function create(bounds: ElementAndBounds[], options: { revision?: number, alignments?: ElementAndAlignment[], requestId?: string }): ComputedBoundsAction {
```

packages/sprotty-protocol/src/actions.ts
```diff
@@ -0,0 +1,500 @@
...
+        return {
+            kind: KIND,
+            bounds,
+            revision: options.revision,
+            alignments: options.alignments,
+            responseId: options.requestId
```
tortmayr, line 293: If the responseId is converted to an optional option we have to default to empty string here
```suggestion
            responseId: options.requestId ?? ''
```

packages/sprotty-protocol/src/actions.ts
```diff
@@ -0,0 +1,500 @@
...
+export interface SetViewportAction extends Action {
+    kind: typeof SetViewportAction.KIND;
+    elementId: string
+    newViewport: Viewport
+    animate: boolean
+}
+export namespace SetViewportAction {
+    export const KIND = 'viewport';
+
+    export function create(newViewport: Viewport, options: { elementId: string, animate?: boolean }): SetViewportAction {
```
tortmayr, line 492: I think, either  `newViewport` should be moved into options here or `element` should be moved out of the options object and options should be made optional (default to empty). But the current split into 1 required parameter and 1 required option object doesn't make sense to me

tortmayr (changes requested):
Thanks! :+1: . Overall the new protocol package looks very good to me. Your proposed changes definitely make sense. I especially like the omitting of the "Schema" suffix. Regarding the potential naming conflict of internal and external model I'd personally opt for option 1:
> Keep it as it is now; in case you need to import both the internal and external model in one file, you can resolve the conflict at the import declaration (this should rarely be necessary in user code).

I noticed a couple of inconsistencies between the implementation of the "old" Request-Response actions and the new interfaces and have some other minor improvements/suggestions.

packages/sprotty-protocol/src/actions.ts
```diff
@@ -0,0 +1,500 @@
...
+}
+
+/**
+ * Sent from the client to the model source (e.g. a DiagramServer) in order to request a model. Usually this
+ * is the first message that is sent to the source, so it is also used to initiate the communication.
+ * The response is a SetModelAction or an UpdateModelAction.
+ */
+export interface RequestModelAction extends RequestAction<SetModelAction> {
+    kind: typeof RequestModelAction.KIND
+    options?: { [key: string]: string | number | boolean }
```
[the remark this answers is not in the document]
spoenemann: I agree, but remember that our current Java server implementation uses `Map<String, String>` as type, you you'll need to be careful to use only primitive values with the Java server.

spoenemann:
@tortmayr I added an action handler mechanism to DiagramServer to promote a function-based approach instead of forcing to subclass DiagramServer even for simple use cases. Could you have a look?

packages/sprotty-protocol/src/actions.ts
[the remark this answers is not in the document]
tortmayr: Ah that's a fair  point. I overlooked that the JsonMap type allows `JsonAny` as values and not  `JsonPrimitive`. In that case it might be better to keep it as it.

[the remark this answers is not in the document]
spoenemann: Using `JsonMap` as you proposed could be useful for NodeJS backends. I already changed it that way and I'd propose to keep that.

tortmayr (approved):
LGTM! :+1:

PR #254, opened 2021-11-17
BezierEdgeRouter improvements

kaisalmen:
Implement review comments:
- Ensure the action is serializable
- Rename action and command
- Moved main execution code to command and removed unnecessary methods

Signed-off-by: Kai Salmen <kai.salmen@typefox.io>

### DIFF
--- a/packages/sprotty/src/features/routing/bezier-edge-router.ts
+++ b/packages/sprotty/src/features/routing/bezier-edge-router.ts
@@ -321,31 +321,31 @@ export class BezierMouseListener extends MouseListener {
         const result = [];
         if (target instanceof SRoutingHandle && (target.kind === 'bezier-add' || target.kind === 'bezier-remove')) {
             if (target.type === 'bezier-create-routing-point') {
-                result.push(new AddRemoveBezierSegmentAction(AddRemoveBezierSegmentActionTask.ADD, target.id));
+                result.push(AddRemoveBezierSegmentAction.create('add', target.id));
             } else if (target.type === 'bezier-remove-routing-point') {
-                result.push(new AddRemoveBezierSegmentAction(AddRemoveBezierSegmentActionTask.REMOVE, target.id));
+                result.push(AddRemoveBezierSegmentAction.create('remove', target.id));
             }
         }
         return result;
     }
 };
 
-export enum AddRemoveBezierSegmentActionTask {
-    ADD,
-    REMOVE
+export interface AddRemoveBezierSegmentAction extends Action {
+    kind: typeof AddRemoveBezierSegmentAction.KIND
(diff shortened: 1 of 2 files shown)


packages/sprotty/src/features/routing/bezier-edge-router.ts
```diff
@@ -317,79 +317,64 @@ export class BezierEdgeRouter extends AbstractEdgeRouter {
...
 };
 
-export enum BezierActionTask {
+export enum AddRemoveBezierSegmentActionTask {
     ADD,
     REMOVE
 }
 
-export class BezierCurveAction implements Action {
+export class AddRemoveBezierSegmentAction implements Action {
```
spoenemann, line 338: Following my recent refactoring, I'd change this to an interface:
```typescript
export interface AddRemoveBezierSegmentAction extends Action {
    kind: typeof AddRemoveBezierSegmentAction.KIND
    targetId: string
    actionTask: 'add' | 'remove'
}
export namespace AddRemoveBezierSegmentAction {
    export const KIND = 'addRemoveBezierSegment';
    export function create(actionTask: 'add' | 'remove', targetId: string): AddRemoveBezierSegmentAction {
        return {
            kind: KIND,
            actionTask,
            targetId
        };
    }
}
```

packages/sprotty/src/features/routing/bezier-edge-router.ts
```diff
@@ -317,79 +317,64 @@ export class BezierEdgeRouter extends AbstractEdgeRouter {
...
-                result.push(new BezierCurveAction(router, BezierActionTask.REMOVE, target.pointIndex, target.parent.id));
+                result.push(new AddRemoveBezierSegmentAction(AddRemoveBezierSegmentActionTask.REMOVE, target.id));
             }
         }
         return result;
     }
 };
 
-export enum BezierActionTask {
+export enum AddRemoveBezierSegmentActionTask {
```
spoenemann, line 333: I'd use a simple string with constrained type: `'add' | 'remove'`

kaisalmen: Yes, constraining a type like this makes enums "obsolete" in this case. 👍

packages/sprotty/src/features/routing/bezier-edge-router.ts
kaisalmen: I will adapt as you suggested. Do you want to refactor all actions implementations in the future?

spoenemann: Yes, actions should always be defined as interfaces.

spoenemann (approved):
:+1:

PR #258, opened 2021-11-23
Improved diagram server API

spoenemann:
This PR improves a few aspects of the new `DiagramServer`:

 * Extracted service definitions to a separate file
 * Made `state` public
 * Made `layoutEngine` optional
 * LayoutAction is executed regardless of the `needServerLayout` option

It also refines the content of the recently copied `sprotty-elk` package:

 * Separated plain definitions and InversifyJS specific definitions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 node_modules/
 npm-debug.log
 yarn-error.log
+.DS_Store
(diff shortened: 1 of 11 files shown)


spoenemann:
I tested the new `sprotty-protocol` and the modified `sprotty-elk` with a Node.js backend that I'll soon publish to `sprotty-vscode` as additional example. I refined some APIs while implementing that example, and finally could verify that everything works 🎉 

@tortmayr, @planger or @NiklasRentzCAU could one of you have a quick look before I finally publish the new version? In particular, I had to remove the Inversify annotations from the `sprotty-elk` definitions to avoid it being included in backend applications that use a different DI library.

spoenemann:
Merging and publishing now. If you find things to improve, we can publish new versions afterwards.

planger:
Sorry, this fell through the cracks. Your change looks great though! Thanks!

PR #260, opened 2021-12-03
Extracted action handler registry from diagram server

spoenemann:
In my first draft of the diagram server, I included the option to register action handlers so it's not necessary to subclass `DiagramServer` just to process incoming actions. But `DiagramServer` instances must be created per client diagram, so there can be possibly many such instances in the backend. I think it would be more convenient to register the action handlers in a separate service instead of the `DiagramServer` instances.

### DIFF
--- a/configs/.eslintrc.js
+++ b/configs/.eslintrc.js
@@ -18,37 +18,7 @@ module.exports = {
     ],
     "rules": {
         "@typescript-eslint/indent": "off",
-        "@typescript-eslint/naming-convention": [
-            "warn",
-                {
-                    "selector": 'default',
-                    "format": ['camelCase'],
-                    "leadingUnderscore": 'allow',
-                    "trailingUnderscore": 'allow',
-                },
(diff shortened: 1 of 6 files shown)


NiklasRentzCAU:
Looks good to me, also in combination to the previous change in [4c111d2a](https://github.com/eclipse/sprotty/commit/4c111d2a96b8e2bfef6acb290ec0d185ddb752c7) to allow to decouple action handling from the diagram server. I think such a change would also be sensible to be implemented in the sprotty-server code to have a consistent API in both systems, wdyt?

spoenemann:
> I think such a change would also be sensible to be implemented in the sprotty-server code to have a consistent API in both systems

We already have a slightly different approach there. Instead of generic action handlers, we offer interfaces for specific purposes: `IPopupModelFactory`, `IDiagramSelectionListener`, `IDiagramExpansionListener`, `IDiagramOpenListener`

Issue #264, opened 2022-01-10
Hover feedback isn't reliably removed


planger:
Steps to reproduce:

![Peek 2022-01-10 17-25](https://user-images.githubusercontent.com/588090/148801294-706e3459-de6f-45d7-aa0e-243f0e21a47a.gif)

As can be seen, the hover feedback -- ie the edge has a wider stroke -- isn't removed when we move the mouse out of the element. It is only removed after enter and leave the element again with the mouse.

On the mouse-out event, the `HoverMouseListener` obtains the next `hoverable` element from the current `target` element and removes the hover feedback by sending a `HoverFeedbackAction` for this element.

```typescript
    mouseOut(target: SModelElement, event: MouseEvent): (Action | Promise<Action>)[] {
        const result: (Action | Promise<Action>)[] = [];
        if (!this.mouseIsDown) {
            ...
             const hoverTarget = findParentByFeature(target, isHoverable);
             if (hoverTarget !== undefined) {
                 result.push(HoverFeedbackAction.create({ mouseoverElement: hoverTarget.id, mouseIsOver: false }));
```

However, if the mouse-out position happens to be above an `SRoutingHandle` the line `findParentByFeature(target, isHoverable)` would return the routing handle and thus, we'd only remove the feedback from the routing handle and not from the edge.

PR #265, opened 2022-01-10
Remove hover feedback for last element

planger:
Remove the hover feedback from the last hover feedback element also
if the mouse-out event occurs for another element. This is needed e.g.
when an edge is entered with the mouse where the routing handle would
appear on selection and then left via the routing handle.

Fixes https://github.com/eclipse/sprotty/issues/264


spoenemann (approved):
Thanks!

Issue #271, opened 2022-01-25
Improve extensibility of elk layout engine


tortmayr:
Currently the `ELKLayoutEngine` (and related concepts) directly call the `getBasicType` utility function to compute the basic type for a given smodel element. This means the behavior of `getBasicType` cannot be overwritten in subclasses. For the GLSP Node server for instance we want to reuse sprotty's `ELKLayoutEngine` but we use a different approach to retrieve the basic type of an SModelElement.

Currently we basically have to duplicate all of the implementation code and replace the usage of the `getBasicType` function manually.
This results in a lot of duplicate code that we also need to update manually if changes to the Sprotty `ELKLayoutEngine` are made.

A simple fix for this would be to refactor the basic type retrieval in a protected method.

PR #272, opened 2022-01-25
#271 Improve extensibility of ElkLayoutEngine

tortmayr:
Refactor usages of the `getBasicType` utility function in `ElkLayoutEngine` and related classes into protected methods. This makes it easier to overwrite the behavior in custom subclasses.


planger (approved):
I don't see any harm in that and it increases extensibility. :+1:

PR #275, opened 2022-02-16
Fix command palette displaying spaces

gabrieldulca:
Signed-off-by: Gabriel Dulca <victor.dulca@tuwien.ac.at>


tortmayr (changes requested):
Thanks, in general the change looks good to me. However, I noticed a weird behavior when no value is entered into the command palette. In this case the regex matching doesn't seem to work and emtpy `<em>` tags are added in between each char<br><br>


Screenshot from 2022-02-16 13-18-08

So while it is displayed correctly the underlying html is technically wrong so I guess we should check  the  given `value` and only apply the em matching if it's not empty.

tortmayr (changes requested):

tortmayr (approved):

PR #280, opened 2022-03-21
Ensure consistent usage of protocol actions

tortmayr:
- Ensures that the new action definitions of `sprotty-protocol` are used consistently over the deprecated class-based definition.
- Ensure that all commands with injected actions use the sprotty-protocol type instead of its deprecated counterpart
- Update deprecated action class definitions so that the also implement the corresponding interface from `sprotty-protocol`. This ensures that the deprecated definitions are compatible on a type level with the new protocol actions. To avoid name clashings the protocol actions are imported with a `Protocol` prefix.
- Also: fix some ESLint warnings

This also fixes problems when extending one of the command classes provided by sprotty.
For instance with the current master state it would not be possible to extend the `UpdateModelCommand` and use
the protocol `UpdateModelAction` as type for the action parameter. (The deprecated variant doesn't allow undefined for the animate property while the new protocol does)
The adopter would be forced to use the deprecated variant to satisfy the typescript compiler.

### DIFF
--- a/packages/sprotty-elk/src/elk-layout.ts
+++ b/packages/sprotty-elk/src/elk-layout.ts
@@ -51,7 +51,7 @@ export class ElkLayoutEngine implements IModelLayoutEngine {
         });
     }
 
-    protected getBasicType(smodel:SModelElement): string{
+    protected getBasicType(smodel: SModelElement): string{
         return getBasicType(smodel);
     }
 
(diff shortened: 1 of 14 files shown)


tortmayr:
There seem to be a lot changes due to formatting. @spoenemann I did use the default TS & JS formatter from VS Code. Are you using something different?

spoenemann (approved):
Thanks!

I usually don't apply auto-formatting, but I'm ok with committing changes made with the default formatting settings.

Issue #287, opened 2022-04-16
Line jump calculation does not work correctly for bezier edges


tortmayr:
When playing around with the class diagram example I noticed that the line jumping calculation for bezier edges doesn't seem to work as expected:
![weird-routing](https://user-images.githubusercontent.com/2311075/163667625-b29ffb20-7913-461c-92ad-1bb4102d723d.gif)

I guess we should try to fix this and in the meantime disable line jumping for bezier edges. IMO it's better do don't have any line jumps instead of wrongly computed ones.

@planger As the original contributor of the line jumping feature,do yo maybe have any additional insights on what could cause this problem and how to fix it?

planger:
Yes, this is because the [`IntersectionFinder`](https://github.com/eclipse/sprotty/blob/master/packages/sprotty/src/features/edge-intersection/intersection-finder.ts) obtains the routing points of an edge to get its segments. With a [sweepline algorithm](https://github.com/eclipse/sprotty/blob/master/packages/sprotty/src/features/edge-intersection/sweepline.ts) it then finds the intersections among the segments. However, this algorithm assumes that the segments are straight lines.

I'm afraid, it'll be quite some work to compute the intersections among or with bezier edges, as it would potentially require a different algorithm that takes the curves into account.

Imho we should document in the [typedoc](https://github.com/eclipse/sprotty/blob/df4577570c5d95356a4ceb3ecc06754e6a099eec/packages/sprotty/src/features/edge-intersection/intersection-finder.ts#L47) that this doesn't work for bezier edges.

spoenemann:
Right, this was already noted in https://github.com/eclipse/sprotty/pull/245#pullrequestreview-775887389 but we forgot to create a separate issue for it.

planger:
The intersection finder ignores bezier edges with 9241c35. Should we consider this task closed, or do we want to keep it open, in case someone would like to work on supporting line gaps and jumps for bezier lines?

PR #290, opened 2022-05-06
Fix edge views that consider intersections

### LINKED ISSUE #277: Jumps at intersections to back-and-forth on east-to-west segments.
I came across a bug in the implementation of the `JumpingPolylineView` which causes intersecting edge paths to jump across one another. Everything's working as intended normally but in my diagram I noticed some lines that seemed strange to me:

image

I noticed that this behavior affected lines going from east to west (the above one in the screenshot). This has to do with how the path is constructed in the `intersectionPath()` method. Intersections are sorted `BY_X_THEN_Y` and if multiple intersections are on that same segment, the end of one intersection connects to the start of the next intersection which causes the line to jump back-and-forth and draw over itself.

To fix this, I extended the `JumpingPolylineView` class, overriding the `intersectionPath` method, to process the intersections in reverse order when the current segment is going the other way:

```    
protected intersectionPath(edge: SEdge, segments: Point[], intersectingPoint: IntersectingRoutedPoint, args?: IViewArgs): string {
        let path = '';
        let intersections = intersectingPoint.intersections.sort(BY_X_THEN_Y);
        const pBefore = segments[segments.indexOf(intersectingPoint) - 1]; // will always be present as the first segment is the starting point
        if (intersectingPoint.x < pBefore.x) {
            // if the segment goes from right to left, we have to add the intersections in different order
            intersections = intersections.reverse();
        }
        for (const intersection of intersections) {
            [...]
```

The rest of the code remains untouched. This bug probably also affects `PolylineEdgeViewWithGapsOnIntersections` but as this uses the same implementation of `intersectionPath` in `JumpingPolylineView` it would also be fixed by this change.

@planger I figure you implemented this feature in the first place. I would love to contribute, but my time is somewhat limited and I'm not so much for creating an Eclipse whatever account to be allowed to open a pull request just for these few lines of code. Also at this point I don't have all the insight to evaluate whether this is the only case where this bug triggers and whether my fix might introduce issues in different scenarios that I'm not seeing right now. In my diagram, I'm only using straight orthogonal edges (north-south, east-west) so there might be stuff going on with slanted lines (although I'm not expecting it). Also I don't know if I'd have to write any tests to ensure this performs properly. 

So if you or anyone else would be so kind to get this bugfix into the library code at some point so others can also benefit from it, I would very much appreciate it. Thank you very much!

Side-note: Using this library has been great so far. The ease of use and design principles used here had me rapidly integrate and modify it to my needs and it works beautifully (I am using it in conjunction with the awesome ELKjs lib).

planger:
Also...
  * Fixes launch config for running single tests
  * Adds line gaps to random-graph example for testing
  * Mention that intersection finder only works for straight segments (#287)

### How to test
I've added line gaps to the random-graph example to simplify verifying whether all intersections are correctly marked.
So I suggest to use that for verifying whether it works now:

https://user-images.githubusercontent.com/588090/167085594-6e062fff-fb10-4d0f-aefd-90e63606f41a.mp4

Fixes #277

Change-Id: I3c72de91188e32fcb296ef81f27c6b03e18c8bd0

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,8 +11,8 @@
             "args": [
                 "${file}",
                 "--no-timeouts",
-                "--opts",
-                "${workspaceRoot}/configs/mocha.opts"
+                "--config",
+                "${workspaceRoot}/configs/.mocharc.json"
             ],
             "env": {
                 "TS_NODE_PROJECT": "${workspaceRoot}/tsconfig.json"
--- a/examples/random-graph/src/di.config.ts
+++ b/examples/random-graph/src/di.config.ts
@@ -19,7 +19,7 @@ import ElkConstructor from 'elkjs/lib/elk.bundled';
 import {
     TYPES, configureViewerOptions, SGraphView, SLabelView, ConsoleLogger, LogLevel,
     loadDefaultModules, LocalModelSource, SNode, SEdge, SLabel, configureModelElement,
-    SGraph, RectangularNodeView, PolylineEdgeView
+    SGraph, RectangularNodeView, edgeIntersectionModule, PolylineEdgeViewWithGapsOnIntersections
 } from 'sprotty';
 import { ElkFactory, ElkLayoutEngine, elkLayoutModule } from 'sprotty-elk/lib/inversify';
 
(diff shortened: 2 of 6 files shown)


spoenemann (approved):
Thanks! Would it work to exclude edges with `routerKind: 'bezier'` from the jump / gap processing?

planger:
> Thanks! Would it work to exclude edges with `routerKind: 'bezier'` from the jump / gap processing?

On a first sight, I didn't find a good place to check for the `routerKind`. However, what was easy is to just ignore routes that contain points, which aren't either `source`, `target` or `linear` (see 1760ca2). I feel that's also quite reliable as this is unlikely to be customized, whereas router and edge kinds are likely to be overwritten by adopters.

With 1760ca2 we avoid including bezier curves (or potentially any routes with non `linear` segments) in the search.

WDYT?

spoenemann (approved):
Yes, looks good :+1:

PR #292, opened 2022-05-11
Use more strict inversify version

dhuebner:
Using ^5.0.1 currently pulls 5.1.1 wich is not compatible with 5.0.1 pulled by the downstream projects. Affect e.g. sprotty-vscode, sprotty-elk

### DIFF
--- a/packages/sprotty-elk/package.json
+++ b/packages/sprotty-elk/package.json
@@ -42,7 +42,7 @@
     "sprotty-protocol": "~0.11.0"
   },
   "optionalDependencies": {
-    "inversify": "^5.0.1"
+    "inversify": "^5.1.1"
   },
   "devDependencies": {
     "@types/chai": "^4.2.22",
--- a/packages/sprotty/package.json
+++ b/packages/sprotty/package.json
@@ -55,7 +55,7 @@
     "@vscode/codicons": "^0.0.25",
     "autocompleter": "^5.1.0",
     "file-saver": "^2.0.2",
-    "inversify": "^5.0.1",
+    "inversify": "^5.1.1",
     "snabbdom": "^3.0.3",
     "sprotty-protocol": "~0.11.0",
     "tinyqueue": "^2.0.3"


spoenemann:
This could be a problem with [sprotty-theia](https://github.com/eclipse/sprotty-theia) because Theia explicitly requires `^5.1.1`:

https://github.com/eclipse-theia/theia/blob/f249b3e93787905d960d7e428f2da0c1796b76d3/packages/core/package.json#L49

We can run into problems if multiple versions of Inversify are around. Can we upgrade everything to `5.1.1`?

dhuebner:
@spoenemann 
I'm okay with updating to 5.1.1, but we need to do it for all the dependency chain.

Regarding 5.1.1, I'm getting following error in yang-vscode main.ts:
```ts
import 'reflect-metadata';
```
```
Module not found: Error: Can't resolve 'reflect-metadata' in '/Users/dhuebner/git/yang-vscode/webview/src'
 @ ./src/main.ts 23:0-27
```

But maybe I'm doing something wrong. However we then need to fix it

dhuebner:
@spoenemann 
I've updated the version to `^5.1.1` like in Theia

spoenemann (approved):

tortmayr:
In the long run maybe we should consider to declare inversify as a `peerDependency` ?
In order for depdency injection to work reliably it is important that all dependent packages resolve the exact same version of inversify. With a direct dependency there is no way to guarantee or control this. Depending on the integration context you might end up with different versions.
For sprotty we have to consider a lot of potential integration cases
- Standalone
- As Theia extension
- As Vscode extension
- As an Eclipse RPC plugin
- Integrated in another web application framework (Angular, nextjs, react ...)
- Indirectly integrated with GLSP's usecases (standalone, theia, vscode, eclipse)
...

This makes it nearly impossible to ensure that exactly one version of inversify is resolved. With using `peerDependencies` we would delegate the responsibility to the sprotty adopter/ the downstream project.

tortmayr:
I have created https://github.com/eclipse/sprotty/issues/296 to track this

PR #303, opened 2022-07-14
Rework ButtonHandler registration

### LINKED ISSUE #300: ButtonHandler API does not work with injection
It looks like with the current implementation of the `IButtonHandler` resp. the `ButtonHandlerRegistry` its not possible to create Button handlers that use injected properties. The `ButtonHandlerRegistry` is directly creating new handler instances using the new keyword hence injected values are not set.

We should probably use real inversify factory injection here (https://github.com/inversify/InversifyJS/blob/master/wiki/factory_injection.md). 

@spoenemann Can we easily change this or is there a specific reason that forces us to use currently implemented approach?

sgraband:
With the current bindings it is not possible to inject into a ButtonHandler.
The registry now recieves a factory instead of a constructor.
Also provide helper method to bin new ButtonHandlers.

Resolves #300

Contributed on behalf of STMicroelectronics

### DIFF
--- a/packages/sprotty/src/base/types.ts
+++ b/packages/sprotty/src/base/types.ts
@@ -24,7 +24,7 @@ export const TYPES = {
     ActionHandlerRegistryProvider: Symbol('ActionHandlerRegistryProvider'),
     IAnchorComputer: Symbol('IAnchor'),
     AnimationFrameSyncer: Symbol('AnimationFrameSyncer'),
-    IButtonHandler: Symbol('IButtonHandler'),
+    IButtonHandlerRegistration: Symbol('IButtonHandlerRegistration'),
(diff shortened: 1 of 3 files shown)


tortmayr:
Code in general looks good to me. I haven't had time to test it though. 
One thing, I think instead of removing we should deprecate the old Buttonhandler registration API. This avoids unnecessary API breaks and conflicts for adopters

sgraband:
Thanks for the feedback!

I readded the constructor injection and deprecated it.

tortmayr (approved):
Thanks @sgraband. Code looks good,
and configuring injectable button handlers now works as expected.

tortmayr:
@spoenemann Any objections? Otherwise I would merge this

planger (approved):
Thanks @sgraband, changes look good to me too!

PR #304, opened 2022-07-14
No Key events when using Projections

### LINKED ISSUE #299: No KeyListeners when using Projections
Using the [Projections](https://github.com/eclipse/sprotty/pull/240) seems to cause the [KeyTool](https://github.com/eclipse/sprotty/blob/master/packages/sprotty/src/base/views/key-tool.ts) to register its events on the `div` element instead of the `svg` element. As a result pressed keys are not registered anymore.

This behavior can be reproduced within the Sprotty Prerendered SVG / HTML Example.

dhuebner:
Enables focus for the root `div` element 

Fix  #299


tortmayr (approved):
Thanks @dhuebner, the fix works like a charm and key listeners now work properly for all views.

PR #308, opened 2022-08-26
Use service binding for all Mouse and KeyListeners

dhuebner:
Make it simpler to extend existing mouse and key listeners.

### DIFF
--- a/examples/classdiagram/src/di.config.ts
+++ b/examples/classdiagram/src/di.config.ts
@@ -46,7 +46,8 @@ export default (containerId: string) => {
         bind(TYPES.ISnapper).to(CenterGridSnapper);
         bind(TYPES.IEditLabelValidator).to(ClassDiagramLabelValidator);
         bind(TYPES.IEditLabelValidationDecorator).to(ClassDiagramLabelValidationDecorator);
-        bind(TYPES.MouseListener).to(BezierMouseListener);
+        bind(BezierMouseListener).toSelf().inSingletonScope();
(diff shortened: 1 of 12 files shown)


tortmayr:
Great change! This definitely makes it easier to customize existing listeners

 > Still need to review which Listeners can be bound in singleton scope

IMO all mouse listeners can  be bound in singleton scope without side effects.
Currently they are only constructor injected into the `MouseTool`. The MouseTool is bound in Singleton scope which effectively makes the injected listeners singleton as well.

dhuebner:
@tortmayr 
> IMO all mouse listeners can be bound in singleton scope without side effects. 

Now all MouseListeners are singletons. I hope I didn't overlooked one...

spoenemann:
@dhuebner would you mind doing the same for KeyListeners?

dhuebner:
@spoenemann 
In the same PR? 
Sure, also all as singleton binding?

tortmayr:
> Now all MouseListeners are singletons. I hope I didn't overlooked one...

@dhuebner Looks good to me. I guess we should also do the same for `PopupMouseListeners`?.

spoenemann:
@dhuebner yes, KeyListeners and PopupMouseListeners behave with the same concept as MouseListeners. It would be great to have a consistent DI approach with all of these.

dhuebner:
@tortmayr @spoenemann 
I think, I have all of them migrated now

spoenemann (approved):

Issue #196, opened 2020-11-21
Alternative context menu handling


avshenuk:
The current logic of context menu handling has certain drawbacks:
- whenever you right-click on an empty area or a new node - it doesn't "deselect" old items same way as it does when you left-click there. And since we use the selected items as the "context" of the menu - you might get undesired results. 
- the temporary select is not rendered right away for some reason - it only shows up after you do a few mouse movements. This probably has something to do with the fact that it doesn't go via the command stack but I'm not sure. 
- the revert of the temp selection is not effective as most of the time you need to left-click somewhere in order to close the context menu and that triggers selection event on its own.

So I suggest an alternative logic by utilizing somewhat similar events issued by the selection listener:
1. If you have something selected already and you right-click on a node FROM that selection - just do nothing and show the context menu. Those selected nodes are your context.
2. If you have something selected already and you right-click on an empty area or a node OUTSIDE that selection - deselect the old nodes AND select the new node (if present). Most of the time that right click means that you don't care about the previous selection - if you do, you can always "Ctrl/Cmd select" all necessary nodes first. 
3. Somewhat debatable: always switch off editing of SRoutableElements whenever you right-click.

A suggestion code that worked quite well for me:
```
export class ContextMenuMouseListener extends MouseListener {
    mouseDown(target: SModelElement, event: MouseEvent): (Action|Promise<Action>)[] {
        const result = [];
        if (event.button === 2) {
            const mousePosition = {x: event.x, y: event.y};
            const selectableTarget = findParentByFeature(target, isSelectable);
            const deselect: SModelElement[] = toArray(target.root.index.all()
                .filter(element => isSelected(element)
                    && !(selectableTarget instanceof SRoutingHandle && element === selectableTarget.parent as SModelElement)));
            if (selectableTarget) {
                if (!selectableTarget.selected) {
                    result.push(new SelectAction([selectableTarget.id], deselect.map(e => e.id)));
                    result.push(new BringToFrontAction([selectableTarget.id]));
                }
            } else {
                result.push(new SelectAction([], deselect.map(e => e.id)));
            }
            const routableDeselect = deselect.filter(e => e instanceof SRoutableElement).map(e => e.id);
            if (routableDeselect.length > 0)
                result.push(new SwitchEditModeAction([], routableDeselect));
            result.push(new ContextMenuAction(mousePosition));
        }
        return result;
    }
}

export class ContextMenuAction implements Action {
    static readonly KIND = "contextMenu";
[...]

PR #316, opened 2022-10-28
Contextmenu open's only the context menu for the item under the curso…

FelixJoehnk:
As in Issue #196 I also found it confusing that the context menu opens for the selected items and not for the items currently under the courser. This should solve this issue by simply checking if the item was selected or not. Also, it currently doesn't do anything if the item under the cursor is not selectable, but this should never be the case since the context menu should only open something under the cursor from my point of view.

best Regards


packages/sprotty/src/features/context-menu/mouse-listener.ts
```diff
@@ -46,14 +47,28 @@ export class ContextMenuMouseListener extends MouseListener {
...
-        const menuItems = await this.menuProvider.getItems(target.root, mousePosition);
-        menuService.show(menuItems, mousePosition, restoreSelection);
+
+        if(isSelectable(target)){
+            if(isSelected(target)){
+                const menuItems = await this.menuProvider.getItems(target.root, mousePosition);
+                menuService.show(menuItems, mousePosition, restoreSelection);
+            }else{
+                const options = {selectedElementsIDs: [id], deselectedElementsIDs: Array.from(root.index.all().filter(isSelected), (val) => {return val.id;})};
+                this.actionDispatcher.dispatch(SelectAction.create(options)).then(() => {
```
spoenemann, line 66: The method is `async`, so you can use `await` here instead of a callback chain.

spoenemann (commented):
Thanks!

Could you fix the eslint warnings (`yarn lint`)?
```
/workspace/sprotty/packages/sprotty/src/features/context-menu/mouse-listener.ts
  60:9   warning  Expected space(s) after "if"     keyword-spacing
  61:13  warning  Expected space(s) after "if"     keyword-spacing
  64:14  warning  Expected space(s) before "else"  keyword-spacing
  64:14  warning  Expected space(s) after "else"   keyword-spacing
```

packages/sprotty/src/features/context-menu/mouse-listener.ts
FelixJoehnk: Hey thanks for the reply. Should all be fine now. 
Best Regards

spoenemann (approved):
:+1:

PR #327, opened 2022-12-12
Small optional type fix in circlegraph example

MazeChaZer:
At this point, the point argument must be optional because the `createContainer` function requires it, and the `addNode` event handler actually calls this function later without arguments.


spoenemann (approved):
Thanks!

PR #328, opened 2022-12-21
Avoid sending label edit action if there is no change

martin-fleck-at:
This problem was observed in [GLSP](https://github.com/eclipse-glsp/glsp/issues/766) as sending an "empty" operation still makes the editor dirty due to a command being on the command stack.


tortmayr (approved):
Thanks! Code looks good and works as expected

PR #329, opened 2022-12-22
Update dependencies and TypeScript target

### LINKED ISSUE #318: update elkjs to 0.8.2
I want to use the model order options for my application. It would be great if elkjs could be updated to use the current version.

### LINKED ISSUE #322: Upgrade ES target for TS compiler
Sprotty is currently targeting ES5 as target for tsc, which is ancient (2009). IMO we should upgrade to ES2017, which brings native async/await support. This would greatly improve the readability and likely also the performance of the generated code.

Compatibility info: https://caniuse.com/?search=es2017

Are there any concerns against this? Anyone using Sprotty on Internet Explorer or Opera Mini?

spoenemann:
- Update TypeScript target version to ES2017 (closes #322)
 - Update snabbdom to 3.5.1
 - Update elkjs to 0.8.2 – this brings breaking changes (closes #318)
 - Update autocompleter to 7.0.1
 - Update dev-dependencies


tortmayr (approved):
I'm not that familiar with sprotty-elk so I don't know how impactful these breaking changes are.
But in general, the changes look good to me and everything still seems to work as expected.
The ES2017 target update should really improve the generated JS code. 👍🏼

PR #330, opened 2022-12-22
SConnectableElement: Add check for index class

### LINKED ISSUE #274: SConnectableElement assumes that the index is a SGraphIndex
SConnectableElement contains an unsafe type cast. We should log a warning in case the index is not a SGraphIndex and give a hint to use SGraph.

spoenemann:
Fixes #274.

Also replaced hidden "alternative" spaces with proper spaces.

### DIFF
--- a/examples/circlegraph/src/views.tsx
+++ b/examples/circlegraph/src/views.tsx
@@ -15,7 +15,7 @@
  ********************************************************************************/
 
 /** @jsx svg */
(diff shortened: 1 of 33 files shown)


spoenemann:
The relevant changes are [here](https://github.com/eclipse/sprotty/pull/330/files#diff-672071b7603010b5466e2d64f529844363c4039718aa08071e652430fc12855f).

dhuebner (approved):
LGTM

PR #332, opened 2023-01-02
Cleanup deprecated types

tortmayr:
Ensure that deprecated types are isolated and don't leak into non-deprecated  definitions. => Consistently use definitions from sprotty protocol in non-deprecated  classes,functions or type definitions.

### DIFF
--- a/packages/sprotty/src/base/model/smodel.ts
+++ b/packages/sprotty/src/base/model/smodel.ts
@@ -15,6 +15,7 @@
  ********************************************************************************/
 
 import { Bounds, isBounds, Point } from "sprotty-protocol/lib/utils/geometry";
+import { SModelElement as ProtocolSModelElement } from "sprotty-protocol";
 import { mapIterable, FluentIterable } from "../../utils/iterable";
 
 /**
@@ -81,8 +82,8 @@ export interface FeatureSet {
     has(feature: symbol): boolean
 }
 
-export function isParent(element: SModelElementSchema | SModelElement):
-        element is SModelElementSchema & { children: SModelElementSchema[] } {
+export function isParent(element: ProtocolSModelElement | SModelElement):
+        element is ProtocolSModelElement & { children: ProtocolSModelElement[] } {
     const children = (element as any).children;
     return children !== undefined && children.constructor === Array;
 }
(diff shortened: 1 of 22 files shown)


packages/sprotty/src/base/ui-extensions/ui-extension.ts
```diff
@@ -26,6 +27,11 @@ export interface IUIExtension {
     id(): string;
     show(root: Readonly<SModelRoot>, ...contextElementIds: string[]): void;
     hide(): void;
+    enableOnStartup?: boolean
+}
+
+export function isUIExtension(object: unknown): object is IUIExtension {
+    return hasOwnProperty(object, 'id') && hasOwnProperty(object, 'show') && hasOwnProperty(object, 'hide');
```
spoenemann, line 34: You could also check that those properties have function type.

packages/sprotty/src/base/model/smodel.ts
```diff
@@ -213,10 +214,10 @@ export function createRandomId(length: number = 8): string {
  * Used to speed up model element lookup by id.
  */
  export interface IModelIndex {
-    add(element: SModelElementSchema): void
-    remove(element: SModelElementSchema): void
-    contains(element: SModelElementSchema): boolean
-    getById(id: string): SModelElementSchema | undefined
+    add(element:ProtocolSModelElement): void
```
spoenemann, line 217: > warning  Expected a space after the ':'  @typescript-eslint/type-annotation-spacing

spoenemann (approved):
Thanks!

This reminds me of a thought: shall we rename the _internal model_ classes such as `SNode`, `SEdge` in order to avoid the name conflict with the protocol interfaces? This name clash is rather confusing for users (where do I need to import that from?)

Example: `SNodeImpl`, `SEdgeImpl` etc.

tortmayr:
> Thanks!
> 
> This reminds me of a thought: shall we rename the _internal model_ classes such as `SNode`, `SEdge` in order to avoid the name conflict with the protocol interfaces? This name clash is rather confusing for users (where do I need to import that from?)
> 
> Example: `SNodeImpl`, `SEdgeImpl` etc.

I aggree, we definitely have to find a cleaner solution than the way we handle it right now. However, I'm not sure if renaming the internal model is the way to go here. This would introduce a major API break and affects a large part of the code base.

We could also go the other way round and rexport sprotty-protocol as part of sprotty with some modifications i.e. renaming the clashing interfaces from sprotty protocol. (e.g. `SModel` -> `SModelSchema`).  This would also mean users can simply import everything directly from `sprotty` and we don't break the existing API.

Pros and Cons for both approaches so we probably should discuss & handle this in a follow-up. 
It might also be time to remove the deprecated action types as they also contribute to this whole name clashing confusion.

spoenemann:
> This would introduce a major API break and affects a large part of the code base.

dhuebner:
@tortmayr 
PR is approved. Should we merge it?

tortmayr:
@spoenemann  I have rebased the change and addressed the two open comments/issues.
We should probably handle the renaming oft the internal model classes with a follow-up PR.

spoenemann (approved):
Thanks!

PR #334, opened 2023-01-10
Fixed edge layout when there are no routing points yet

spoenemann:
When the ELK server is invoked with a graph that does not have any layout information, the edge routes were not applied. This PR fixes that bug.


dhuebner (approved):
Looks good.
Didn't check the launch configs thought

spoenemann:
The launch config changes were due to warnings shown by vscode.

Issue #335, opened 2023-01-23
v0.13.0 Regression: Repeated reconnections of edges deterministically break dragging


MazeChaZer:
How to reproduce:

```
yarn
yarn --cwd examples start
```

Navigate to <http://localhost:8080/classdiagram/class-diagram.html>, choose any edge, reconnect their source or target. The first time it works, if you try to reconnect the source/target again, dragging is not possible anymore at that anchor point and the following JavaScript error is thrown:

```
sprotty CommandStack: Failed to execute command: Error: Duplicate ID in model: edge0_dangling-source
```

It seems like the reconnection of edges leaves `dangling-source` nodes behind that should be cleaned up. The second time this `dangling-source` node already exists and triggers the duplicate ID error.

In v0.12.0 this problem did not exist yet. Looking at the commit history between v0.12.0 and v0.13.0, the commit 0d525e336d7d91e3b75434a5f91b4e3fff77de0f seems suspicious, as it changes exactly this feature.

dhuebner:
Fixed in master

PR #337, opened 2023-02-08
Reconnecting of edges breaks dragging #335

dhuebner:
Fixes ` Error: Duplicate ID in model: edge0_dangling-source`, see #335

How to test: Open the Class diagram example. You should be able to reconnect edges more than once.


tortmayr (approved):
LGTM! 👍🏼  Can confirm that multiple reconnects now work as expected

Issue #310, opened 2022-09-09
Cannot move single routing point with Manhattan routing


martin-fleck-at:
Investigating the [bug reported in GLSP](https://github.com/eclipse-glsp/glsp/issues/657), I saw that there is a bug in the Manhattan routing that prevents moving the single routing point.

![cannot-move-first-point](https://user-images.githubusercontent.com/19170971/189323603-15fe53fa-6cc7-43c3-b9b6-b29c9890f22a.gif)
(it can also be reproduced with the class diagram of Sprotty)

When moving the routing handles (which are placed between two routing points or the start/end point), we want to align the calculated route with the routing points present in the edge. However, if we only have start and end point, then the routing points array of the edge is empty preventing any update.

We can avoid the problem when we simply add a routing point to the edge if we try to move the routing handle. However, it is not very clear to me which routing points would be suitable as it very much depends on the start/end position calculation of the edge. If we simply were to use the target position then we would get something like this:

![fix-move-first-point](https://user-images.githubusercontent.com/19170971/189324417-b1270647-3827-40d4-a116-465dfa5d0ad9.gif)

But this does not behave like we would expect as you can see if I have already an arch between source and target then I get a much nicer move.

rsoika:
@martin-fleck-at  can you show how you did realize your workaround? Currently I am trying to implement my own manhantan router and this would possible help me to understand the inner logic.
@spoenemann it's very hard for me to contribute to your project because the comments are missing from your code. Even if you understand your own implementation, it's not easy for an outsider to get into the code. Routing is generally a complex topic and missing comments do not reduce it ;-)

martin-fleck-at:
@rsoika I opened a WIP PR on the GLSP side to show how I did it. The changed code is basically the [if-block starting on line 33](https://github.com/eclipse-glsp/glsp-client/pull/200/files#diff-030a41434d98fee751a7a15e61b1f29636b694b88e8c653c7a067e483230c0c8R33). Hopefully, this will help you with your own Manhattan router.

PR #338, opened 2023-02-08
Move single routing point with Manhattan #310

dhuebner:
Allows moving single routing point with Manhattan routing, see #310


spoenemann (approved):
Thanks!

Issue #339, opened 2023-02-09
Make it possible to scroll the viewport using alt+wheel


dhuebner:
It would be nice to be able to scroll the viewport using alt + wheel. 
Wheel should still be used to zoom in/out.

dhuebner:
Added to master

PR #340, opened 2023-02-09
Allows scrolling the viewport using alt + wheel

dhuebner:
To try it out open the SVG example, hold `alt` and use mouse wheel or touchpad to scroll.

See #339

### DIFF
--- a/packages/sprotty/src/features/viewport/zoom.ts
+++ b/packages/sprotty/src/features/viewport/zoom.ts
@@ -17,11 +17,11 @@
 import { Action, SetViewportAction } from "sprotty-protocol/lib/actions";
 import { Viewport } from "sprotty-protocol/lib/model";
 import { Point } from "sprotty-protocol/lib/utils/geometry";
-import { getWindowScroll } from "../../utils/browser";
 import { SModelElement, SModelRoot } from "../../base/model/smodel";
-import { MouseListener } from "../../base/views/mouse-tool";
 import { SModelExtension } from "../../base/model/smodel-extension";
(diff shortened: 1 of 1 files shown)


packages/sprotty/src/features/viewport/zoom.ts
```diff
@@ -51,20 +51,32 @@ export class ZoomMouseListener extends MouseListener {
...
-                    y: viewport.scroll.y - offsetFactor * viewportOffset.y
-                },
-                zoom: viewport.zoom * newZoom
-            };
-            return [SetViewportAction.create(viewport.id, newViewport, { animate: false })];
+        if (!viewport) {
+            return [];
         }
-        return [];
+        const newViewport = () => {
```
spoenemann, line 57: Why is this a function? It's executed immediately before the action is created.

spoenemann (commented):
Error:
```
src/features/viewport/zoom.ts(18,1): error TS6133: 'Viewport' is declared but its value is never read.
```

packages/sprotty/src/features/viewport/zoom.ts
dhuebner: It's not a function, it's a `newViewport` supplier! :)
It looked ugly when using return in if/else branches.  
I will have a closer look soon, maybe extract some code to a protected  method and refactor existing code a bit

dhuebner: Extracted zoom and scroll viewport calculation in separate functions for easier customization.

PR #343, opened 2023-02-19
specify host as 127.0.0.1

loganmay:
When running the random graph generator example, I see an error like:
Screen Shot 2023-02-18 at 8 36 23 PM

I noticed the host name looked odd. I found [this SO post](https://stackoverflow.com/questions/10643965/econnrefused-error-with-node-js-that-does-not-occur-in-other-clients) and [this SO post](https://stackoverflow.com/questions/73843720/connect-econnrefused-127017) that explain that node has issues with `localhost` and to use `127.0.0.1` instead. So, I made this change locally to the socket connect call and it fixed the issue:
Screen Shot 2023-02-18 at 8 37 40 PM

### DIFF
--- a/packages/sprotty-elk/src/node/socket-server.ts
+++ b/packages/sprotty-elk/src/node/socket-server.ts
@@ -97,7 +97,7 @@ export class SocketElkServer implements ELK {
                 this.socket = undefined;
                 reject(err);
             });
-            socket.connect(this.port);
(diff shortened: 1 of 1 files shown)


spoenemann (approved):
Thanks! Could you sign the ECA as described in https://github.com/eclipse-sprotty/sprotty/pull/344#issuecomment-1443413649 ?

PR #344, opened 2023-02-19
Add separate websocket example

loganmay:
Switch distributed example to websockets


spoenemann:
Thanks for the contribution @loganmay! I propose to replace the previous HTTP-based solution because it's very restictive. Could you change the existing example to use your WebSocket-based code?

You need to sign the [Eclipse Contributor Agreement](https://accounts.eclipse.org/user/eca) and sign-off your commit (with `git commit -s`) using the same email address as with your Eclipse account to pass the ECA check.

loganmay:
@spoenemann replaced the previous solution, signed the agreement, and signed-off the commit

examples/package.json
```diff
@@ -1,13 +1,14 @@
 {
   "name": "sprotty-examples",
   "version": "0.13.0",
-  "private": "true",
```
spoenemann, line 4: Why did you remove this? We haven't published the examples to the npm registry yet.

examples/server/server-app.ts
```diff
@@ -20,51 +20,46 @@ import { ElkFactory, ElkLayoutEngine } from 'sprotty-elk/lib/elk-layout';
...
+wsServer.on('connection', socket => {
+    socket.on('error', console.error);
+    socket.on('message', message => {
+        try {
+            const incomingMessage = JSON.parse(message.toString()) as ActionMessage;
+            // Create a diagram server and process the incoming message
+            const diagramServer = new DiagramServer(async (action: Action) => {
+                const msg = JSON.stringify({ clientId: incomingMessage.clientId, action });
+                socket.send(msg);
+            }, services);
```
spoenemann, line 46: Proposal: define the DiagramServer outside the message handler (but inside the connection handler). That way it stays the same server instance as long as the connection remains open, which is relevant in more complex cases where the server might send updates to the client.

examples/package.json
loganmay: It happened automatically when I ran `yarn install` oddly enough. I will add it back

spoenemann (approved):
The WebSocket URL needs to be fixed to run in Gitpod, but I'll do that afterwards.

Issue #345, opened 2023-02-22
Context Menu on Label/Button is not shown


BjBe82:
Last working Version: 0.12.0
Bugged Version: 0.13.0

The context menu is no longer triggered (IContextMenuService is not called) if the user right-clicks a label (or button). For smaller nodes where the label covers most of the node, this makes it nearly impossible to open the context menu.

dhuebner:
@BjBe82 
> Last working Version: 0.12.0
> Bugged Version: 0.12.0

Do you mean "Bugged Version: 0.13.0"?

dhuebner:
Probably caused by https://github.com/eclipse-sprotty/sprotty/pull/316

BjBe82:
Yes, I meant 0.13.0 -> update the ticket.

dhuebner:
@BjBe82 
I don't really understand the expected behavior looking into the current implementation. I'm about to create a reproducible example in one of our example projects to understand what and when should happen. But I think the `isSelectable(target)` is what prevents your app to show a context menu. 
https://github.com/eclipse-sprotty/sprotty/blob/master/packages/sprotty/src/features/context-menu/mouse-listener.ts#L60

dhuebner:
@BjBe82 
You can try the current development state in the class diagram example here: https://gitpod.io/#https://github.com/eclipse-sprotty/sprotty/pull/346
Note that the node label itself is not selectable.

Feedback regarding the right-click selection behavior is very welcome here  #346

spoenemann:
Shall this be kept open, or is it fixed with #346?

dhuebner:
@spoenemann 
I think it is fixed now. Probably just forgot to close it

BjBe82:
Sorry for the late reply, I checked this with version 1.0.0 today and can confirm that it is fixed.

PR #346, opened 2023-02-24
Allow context menu on not selectable elements #345

dhuebner:
As in a lot of cases the target element must not be selectable I removed the if condition from `menu-listener.ts`.
- In case of a non selectable element the next selectable parent will be checked.
- In case of root (normally not selectable) the menu will be shown too

I also added a context menu to our class diagram example so we can discuss the current selection behavior.
Please note that a lot of actions act on elements that they retrieve from the available selection tracking services (e.g. GetSelectionAction or LocalModelSource.getSelection).

### DIFF
--- a/examples/classdiagram/class-diagram.html
+++ b/examples/classdiagram/class-diagram.html
@@ -26,7 +26,7 @@
                 <div id="sprotty" class="sprotty"/>
             </div>
             <div class="copyright">
-                &copy; 2021 <a href="https://www.typefox.io/">TypeFox GmbH</a>.
+                &copy; 2023 <a href="https://www.typefox.io/">TypeFox GmbH</a>.
             </div>
         </div>
     </div>
--- a/examples/classdiagram/css/diagram.css
+++ b/examples/classdiagram/css/diagram.css
@@ -88,37 +88,37 @@
     stroke-width: 3px;
 }
 
-.sprotty-edge > .sprotty-routing-handle {
+.sprotty-edge>.sprotty-routing-handle {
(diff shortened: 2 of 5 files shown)


BjBe82:
I had a first look at this, and this behaves in general as i would have expected (see #345). One thing to note (which might be unrelated to the changes here), if i select "Foo" and then right click  "Baz" -> "Deleted Selected" ... both ("Foo" and "Baz") are deleted.

dhuebner:
>  if i select "Foo" and then right click "Baz" -> "Deleted Selected" ... both ("Foo" and "Baz") 

That is exactly the problem I have with the current behavior. Comparing with other common tools like e.g. Finder (mac osx file manager) there is an additional selection marker (border) that shows for which elements the context menu is currently responsible. Imagine you have selected some files (background will change to selection color) and than right click on a non selected file (border in selection color will appear), then the context menu action will only handle the file under cursor and not all the selected. We do not have this kind of selection in Sprotty currently, so I don't know how to handle it properly.

dhuebner:
@spoenemann 
I think we should start a discussion about how the element under the cursor should be handled when processing right-click. 
When any other elements are selected it is pretty clear, we should select it. But when other elements are selected and user right-click over an other (selectable/not-selectable?) element, what should happen? 

However, the issue for this PR is fixed and can be merged :)

spoenemann (approved):
I think it's ok like this. The context menu actions clearly state that they refer to the selection, and IMO it's fine to assume that the right-clicked element is part of it.

PR #347, opened 2023-03-01
added zoomScale property to centerAction to set zoom level when executing the action

jonah-iden:
Added a `zoomScale` property to the `CenterAction` to explicitly set te zoom level when centering the view.

**how to test**
in the circlegraph example:
- add following new button to the `circlegraph.html` file
``` <button id="centerZoomed">center and zoom selection</button>```
-  add following code to the `standalone.ts` to execute the center action on click
```    
document.getElementById('centerZoomed')!.addEventListener('click', async () => {
    dispatcher.dispatch(CenterAction.create(graph.children.map(el => el.id), {animate: true, zoomScale: 2}))
});
```
now when clicking the button the center action is executed and zoom level is set to 2.

### DIFF
--- a/packages/sprotty-protocol/src/actions.ts
+++ b/packages/sprotty-protocol/src/actions.ts
@@ -463,16 +463,18 @@ export interface CenterAction extends Action {
     elementIds: string[]
     animate: boolean
     retainZoom: boolean
+    zoomScale?: number
 }
 export namespace CenterAction {
     export const KIND = 'center';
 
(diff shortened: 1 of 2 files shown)


dhuebner:
@jonah-iden 
Thanks! It works very well.
Btw. there are key bindings for center: `'KeyC', 'ctrlCmd', 'shift'` and focus: `'KeyF', 'ctrlCmd', 'shift'`, maybe we can show them in the circles example. What do you think if we print this short cuts in the right bottom in the same color as © 2017 ? We should probably also change 2017 to © 2017-2023 or just © 2023

dhuebner (approved):

PR #348, opened 2023-03-13
Fixed WS connection for Gitpod

spoenemann:
Follow-up of #344


tortmayr (approved):
LGTM! 👍🏼  The random graph example now works as expected in Gitpod

PR #349, opened 2023-03-15
Improve extensibility of multi-injected keys

### LINKED ISSUE #239: Improve extensibility of multi-injected keys
Double check the codebase and ensure that all all multi-injected classes are bound to self() in Singleton scope before binding them to the multi injection key. This makes it possible to overwrite them.
e.g
```typescript
bind(MoveMouseListener).toSelf().inSingletonScope();
bind(TYPES.MouseListener).toService(MoveMouseListener); 
```
instead of 
```typescript
 bind(TYPES.MouseListener).to(MoveMouseListener); 
```

See also discussion in https://github.com/eclipse/sprotty/pull/234#issuecomment-897606251

tortmayr:
Double check the codebase and ensure that all all multi-injected classes are bound to self() in Singleton scope before binding them to the multi injection key 
Fixes #239


dhuebner (approved):
I think it's a good idea. We were concentrated on mouse and key listener, but all multibindings should be handled this way in my opinion. 
Thanks!

PR #351, opened 2023-04-18
Updated scaledFont to account for decimals

david-pa:
The previous implementation was breaking with decimal numbers, which I talked about [in this discussion post](https://github.com/eclipse-glsp/glsp/discussions/971), so this is a potential fix to prevent this behavior in the future.

### DIFF
--- a/packages/sprotty/src/features/edit/edit-label-ui.ts
+++ b/packages/sprotty/src/features/edit/edit-label-ui.ts
@@ -269,7 +269,7 @@ function getEditableLabels(contextElementIds: string[], root: Readonly<SModelRoo
 }
 
 function scaledFont(font: string, zoom: number): string {
-    return font.replace(/([0-9]+[.]?[0-9]*)/, (match) => {
+    return font.replace(/\d+(\.\d+)?/, (match) => {
         return String(Number.parseInt(match, 10) * zoom);
     });
 }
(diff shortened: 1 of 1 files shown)


packages/sprotty/src/features/edit/edit-label-ui.ts
```diff
@@ -269,7 +269,7 @@ function getEditableLabels(contextElementIds: string[], root: Readonly<SModelRoo
 }
 
 function scaledFont(font: string, zoom: number): string {
-    return font.replace(/([0-9]+)/, (match) => {
+    return font.replace(/([0-9]+[.]?[0-9]*)/, (match) => {
```
planger, line 272:
```suggestion
    return font.replace(/\d+(\.\d+)?/, (match) => {
```

How about this would that work as well? I think your expression would also match '1.'.

planger (commented):
Thank you very much for your contribution!
Please make sure to also sign the [ECA](https://www.eclipse.org/legal/ECA.php) with the email address that you've used for your commit.

I made a suggestion inline to improve the regular expression. Can you please test if that works for your use case too?

packages/sprotty/src/features/edit/edit-label-ui.ts
david-pa: Good point! I'll use this one instead.

planger:
The ECA check is still red, since you committed with an anonymous Git user. But anyway, since it is just such a small change I've just opened another [PR](https://github.com/eclipse-sprotty/sprotty/pull/352).
Thanks again for raising this issue!

PR #352, opened 2023-04-19
Fix font scaling in edit label UI extension

planger:
If the font size property is a number with decimals, the regular expression didn't account for the decimal place. This change enhances the regular expression to match also the numbers after the decimal point.

Initially raised in https://github.com/eclipse-sprotty/sprotty/pull/351

Also this change fixes the VS Code build task of this repository.

Change-Id: I9592518e533e5636e4f3a27319cbe96a5ec71ae1

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -46,7 +46,7 @@
         {
             "label": "Build Sprotty examples",
             "type": "shell",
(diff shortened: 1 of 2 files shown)


tortmayr (approved):
Looks good to me! 👍🏼

PR #354, opened 2023-05-12
Corrected edge label placement for side='on'

### LINKED ISSUE #353: Label placement wrong with no Rotate and edgePlacement 'on'
In the decoration flow for a label with EdgePlacement rotate = false and side = 'on', implemented by EdgeLayoutPostprocessor
- method getAlignement calculates where to draw the label with respect to a point on the edge
- in case of side='on', we move the label with the corner of the bounds to the middle of the bounds (OK)
- for this, we need to add to x bounds.witdth (tobe=ok) and not bound.height (asis=nok).

See line https://github.com/eclipse-sprotty/sprotty/blob/c120cb7de3137e3cf0659e383ba0cd1a6dde4a88/packages/sprotty/src/features/edge-layout/edge-layout.ts#L132

spoenemann:
Fixes #353.


tortmayr (approved):
Looks good to me! 👍🏼

Issue #231, opened 2021-08-11
Graduate to version 1.0?


spoenemann:
I'm thinking about graduating from the incubation phase. I'd like to gather feedback about changes in the code that you would like to see before such a step. This is particularly important for breaking changes: better do them now than after a 1.0 release.

Things I would like to do before 1.0:
 * Finish and merge migration to snabbdom 3 (#227)
 * Consider reducing the number of repositories (create a monorepo)
 * Review the API and data structures

planger:
Thanks for your initiative! I'm all for it and happy to support step.

I'd find it great if we could get #226 too. Also @Aksem is working on an integration of [libavoid](https://www.adaptagrams.org/documentation/libavoid.html) with Sprotty via a webassembly port of that library. This integration should probably live in an optional separate package, but it requires a few minor additional changes to the router API on top of #226, which imho would be good to have in an 1.0 release.

CamilleLetavernier:
Hi,
I've created a new issue (#232), which may also slightly affect the API, and would probably make sense before the 1.0 release. I will provide a PR for that

tortmayr:
Sounds like a great idea! :+1: 
API-wise I think there are a couple of points we should discuss before the 1.0 release.

- In PR #226 Miro raised the question wehter we should refactor the action API to use interfaces instead of classes. I opened #233 to track this. Even if we decide to not move away from the class-based approach we should at least ensure that all action have a proper guard function and use them over `instance of` checks.
- In #33 we introduced a `Tool (manager) API` to support enabling and disabling tools (i.e. mouse listeners) at runtime. This API is heavily used in GLSP but currently not used in sprotty at all. We should discuss how to go forward with this. Is the `Tool API` still generally interersting for sprotty? If so, should we start to migrate the existing tools (e..g 'MoveMouseListener')? If not, maybe we should move this API from sprotty directly to GLSP.
- When introducing the `UIExtension API` (#63) the question was raised wether we should refactor the Popup stuff to an UI-extension based approach (https://github.com/eclipse/sprotty/pull/63#pullrequestreview-205640214). If we still want to do that we should probably also do it before the 1.0 release.

spoenemann:
Thanks @tortmayr! Could you create issues to track discussions about the second and third point as well?

I think both points would require an analysis of feasibility and consequences of a refactoring.

spoenemann:
[...]

PR #355, opened 2023-05-16
Rename internal model classes with 'Impl' suffix

spoenemann:
This has been discussed before for quite some time, now I'd like to make a step forward in preparation of #231: The _internal model_ classes like `SModelElement`, `SNode` etc. have the same names as the _external model_ interfaces defined in `sprotty-protocol`. This is a source of confusion for users.

Historically, the external model interfaces had a `Schema` suffix to disambiguate, but I removed that during the move to the `sprotty-protocol` package. I now regard the external model as the primary interface. The internal model is used primarily at the diagram configuration with InversifyJS (`configureModelElement`) and as arguments to custom services such as views.

I propose to use the `Impl` suffix for internal model classes and deprecate the previous symbols. I would release v0.14.0 with this change (and maybe others) so users can migrate gracefully with the deprecation notices. Then I'd remove all deprecated code when we shift to v1.0.0.

### DIFF
--- a/examples/circlegraph/src/di.config.ts
+++ b/examples/circlegraph/src/di.config.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2017-2018 TypeFox and others.
+ * Copyright (c) 2017-2023 TypeFox and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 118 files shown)


dhuebner (approved):
I think the `Impl` suffix fits well and makes things clear.
 
Review all the rename, copyright and quotation mark changes is not really necessary I suppose.

tortmayr (approved):
LGTM! 👍🏼

PR #356, opened 2023-05-17
Allow limits on zooming and panning

### LINKED ISSUE #10: Allow limits on zooming / panning
From theia-ide/sprotty#220:

> One can get lost by zooming extremely far in or out or by panning far away. It should be possible to configure limits on how far one can zoom and pan.

spoenemann:
Closes #10. The new feature can be tested with the SVG example

Shall we provide sensible defaults for the zoom factor? Zooming towards infinity doesn't make a lot of sense.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,12 +6,10 @@
             "type": "node",
             "request": "launch",
             "cwd": "${workspaceRoot}",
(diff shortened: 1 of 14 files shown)


spoenemann:
I made a proposal for default limits, see e.g. the generated graph example to test.

dhuebner (approved):
Works good! Especially zoom level is very helpful

PR #357, opened 2023-05-17
Bump TypeScript 5 and Inversify 6

kaisalmen:
When upgrading to TypeSript 5 and using `@inject` in a constructor the following error is raised:
```
Argument of type 'undefined' is not assignable to parameter of type 'string'
```
Upgrading to inversify 6 resolves the issue.

See this comment: https://github.com/microsoft/TypeScript/issues/52435#issuecomment-1405483291 and the following.

Related: https://github.com/eclipse-theia/theia/pull/12425

Inversify is now a `peerDependency` addressing: https://github.com/eclipse-sprotty/sprotty/issues/296

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -20,7 +20,7 @@
     "source-map-loader": "^4.0.1",
     "style-loader": "^3.3.1",
     "ts-loader": "^9.4.2",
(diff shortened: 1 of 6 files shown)


spoenemann (approved):
Thanks!

PR #359, opened 2023-05-25
Fix mocha launch configurations

tortmayr:
Updates outdated launch configs for mocha tests to correctly use `ts-mocha`


spoenemann (approved):
Thanks!

PR #360, opened 2023-05-25
Clean up codebase with regards to deprecated API

tortmayr:
Ensure that deprecated Action & SModel API is encapsulated and do not leak into non-deprecated API by
- Migrating the examples to the new Action/SModel API
- Migrating the test cases to the new Action/SModel API
- Remove any remaining usage of deprecated Action/SModel API in the non-deprecated code base

In addition, this change restructures all files that contain deprecated Action/SModel API and moves the deprecated definitions in a separate section on the end of the file.

This enables easy removal of the deprecated API with the 1.0.0 release. The deprecation section at the file end can be safely removed without breaking codebase.


spoenemann (approved):
Very helpful cleanup, thank you!

PR #361, opened 2023-05-29
Track promise  state in Deferred

tortmayr:
Keep track of the current promise state in a Deferred. Offers the possibility to query the current promise state via a sync property.


spoenemann (approved):
Looks like a useful change 👍

PR #363, opened 2023-06-02
Add default layoutableChildFeature to SNodeImpl

gfontorbe:
Adds the default feature `layoutableChildFeature` to `SNodeImpl` so that children of an `SNodeImpl` follow the node's layoutOptions by default.


spoenemann:
Could you describe in which scenario this change is relevant? Looking at the code, it seems the `layoutableChildFeature` is only checked for direct children of a `LayoutContainer` element (`SNode` or `SCompartment`). Are you using an `SNode` within such a layout container?

gfontorbe:
I think introducing a new component might be more confusing.

I'm okay with leaving this feature out of `SNode`, but I think that deserves some clear explanation in the documentation in favor of using `SCompartment` for micro-layouting.

I will work on updating the documentation when I get time

PR #364, opened 2023-06-02
Fix computation of bounding box with children rendered outside

### LINKED ISSUE #165: Incorrect layouting of ports with labels
When adding a label to a port, the size of that port is changed depending on the label:

port-size
with `PortLabelPlacement.INSIDE`:
port-size-inside

I'm not sure if this is a problem on the server- or client-side. I seems as if the bounds calculation of the port considers the label (since it is a child of the port), while ELK regards these as being indepentent.

To be able to reproduce I've made these changes to the circlegraph example:

`build.gradle`:

```gradle
compile "org.eclipse.elk:org.eclipse.elk.alg.layered:${versions.elk}"
```

`di.config.ts`:

```typescript
configureModelElement(context, 'graph', SGraph, SGraphView);
configureModelElement(context, 'label', SLabel, SLabelView);
configureModelElement(context, 'node', RectangularNode, RectangularNodeView);
configureModelElement(context, 'port', RectangularPort, RectangularNodeView);
configureViewerOptions(context, {
    needsClientLayout: true,
    needsServerLayout: true
});
```

`GraphLayoutEngine.java`:

```java
ElkLayoutEngine.initialize(new LayeredOptions());

configurator.configureByType("graph")
		.setProperty(CoreOptions.ALGORITHM, LayeredOptions.ALGORITHM_ID);
configurator.configureByType("node")
		// .setProperty(CoreOptions.PORT_LABELS_PLACEMENT, PortLabelPlacement.INSIDE)
		.setProperty(CoreOptions.PORT_CONSTRAINTS, PortConstraints.FIXED_ORDER)
		.setProperty(CoreOptions.PORT_ALIGNMENT_DEFAULT, PortAlignment.JUSTIFIED)
		.setProperty(CoreOptions.NODE_SIZE_CONSTRAINTS, SizeConstraint.free());
```

`GraphGenerator.java`:

```java
public SGraph generateGraph() {
	return new SGraph(graph -> {
		graph.setId("graph");
		graph.setChildren(new ArrayList<>());
		graph.getChildren().add(new SNode(node -> {
			node.setChildren(new ArrayList<>(4));
			for (int p = 0; p < 4; p++)
				node.getChildren().add(generatePort(p, "test" + Strings.repeat("_test", p)));
		}));
	});
}

private SPort generatePort(int p, String pLabel) {
	return new SPort(port -> {
		port.setId("port" + p);
		port.setChildren(Collections.singletonList(new SLabel(label -> {
			label.setId("port" + p + "/label");
			label.setText(pLabel);
		})));
	});
}
```

spoenemann:
Fixes #165.

I tried to find a generalization for the fix that would enable us to cover many use cases. The approach is to set a specific attribute to the SVG element that you'd like to use to determine the bounding box. Example:

https://github.com/eclipse-sprotty/sprotty/blob/9ff65633561a7c459e6e3662f0fabd45714f9182/examples/random-graph/src/views.tsx#L31-L40

I added ports with labels to the "Random Graph" example to show this.

Before the fix:
Screenshot 2023-06-02 at 11 39 31

After the fix:
Screenshot 2023-06-02 at 12 16 16

### DIFF
--- a/examples/random-graph/css/diagram.css
+++ b/examples/random-graph/css/diagram.css
@@ -28,8 +28,20 @@
     stroke-width: 3;
 }
 
+.sprotty-port {
+    fill: #484;
+}
+
 .sprotty-edge {
     fill: none;
     stroke: #152;
     stroke-width: 1;
+}
+
+.sprotty-label.node {
+    font-size: 1.2rem;
+}
+
+.sprotty-label.port {
+    font-size: 0.8rem;
 }
--- a/examples/random-graph/src/di.config.ts
+++ b/examples/random-graph/src/di.config.ts
@@ -16,12 +16,18 @@
 
 import { Container, ContainerModule } from 'inversify';
 import ElkConstructor from 'elkjs/lib/elk.bundled';
+import { LayoutOptions } from 'elkjs/lib/elk-api';
 import {
     TYPES, configureViewerOptions, SGraphView, SLabelView, ConsoleLogger, LogLevel,
     loadDefaultModules, LocalModelSource, SNodeImpl, SEdgeImpl, SLabelImpl, configureModelElement,
-    SGraphImpl, RectangularNodeView, edgeIntersectionModule, PolylineEdgeViewWithGapsOnIntersections
+    SGraphImpl, RectangularNodeView, edgeIntersectionModule, PolylineEdgeViewWithGapsOnIntersections,
+    SPortImpl
 } from 'sprotty';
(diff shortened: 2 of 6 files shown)


spoenemann:
@planger @tortmayr @dhuebner any objections to merging this?

planger (approved):
Thanks, looks good to me! :+1: I think this additional attribute may be useful for several use cases.

PR #365, opened 2023-06-08
Extracted `model` getter to abstract ModelSource

spoenemann:
I noticed quite often that the implementation of certain custom functionality requires access to the external model. This change makes the external model available in the `ModelSource` base class by exposing the `model` getter property. Also some code cleanup and commenting on the way.


dhuebner (approved):
Having access to `get model(): SModelRootSchema` is really useful.

PR #371, opened 2023-08-09
Deprecate tool manager API

### LINKED ISSUE #235: Refactor sprotty tools to `ToolManager API`?
In #33 we introduced a Tool (manager) API to support enabling and disabling tools (i.e. mouse listeners) at runtime. This API is heavily used in GLSP but currently not used in sprotty at all. We should discuss how to go forward with this. Is the Tool API still generally interesting for sprotty? If so, should we start to migrate the existing tools (e..g 'MoveMouseListener')? If not, maybe we should move this API from sprotty directly to GLSP.

Requires an analysis of feasibility and consequences of a refactoring.

tortmayr:
Deprecate tool manager API as the entire concept is unused in sprotty and should rather be contributed directly by GLSP. Remove related bindings from default module.
Closes #235


spoenemann (approved):
👍

PR #372, opened 2023-08-22
Fix command palette suggestion position

tortmayr:
Fix the position of the command palette suggestion box by setting the position to "fixed". This seems to be broken since the version bump form commander to version 7.0.1 (https://github.com/eclipse-sprotty/sprotty/pull/329).

Current Master:

Screenshot from 2023-08-22 12-05-25

This PR:
Screenshot from 2023-08-22 12-05-35


planger (approved):
Thanks, looks good to me! :+1:

PR #373, opened 2023-08-25
Layout configuration for the random graph example

dhuebner:
Now the use can choose between Left/Right/Top/Down directions.
Another tweak demonstrates how to keep a selected node in place when the layout is changing. To test, select a node and change the layout direction. Selected node should remain in the same place.

### DIFF
--- a/examples/random-graph/random-graph.html
+++ b/examples/random-graph/random-graph.html
@@ -14,6 +14,15 @@
         <div class="row" id="sprotty-app" data-app="random-graph">
             <div class="col-md-10">
                 <h1>Sprotty Random Graph Example</h1>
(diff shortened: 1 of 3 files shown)


planger (commented):
Very nice addition!

But shouldn't `TrackSelectedUpdateModelCommand` be part of the Sprotty library and not just part of the examples so that it could be reused by adopters?

Thank you!

dhuebner:
@planger 
Yeah, maybe it is a good idea! We could offer this implementation in addition to the default one. But then it needs a better name  :) 
I'm also working on making it a bit smarter. In case there is no selected element, the most centered element should be used to stay in focus.

PR #374, opened 2023-09-01
Removed deprecated API

spoenemann:
In preparation of the v1.0.0 release, we want to remove all API that has been deprecated in the past.


tortmayr (approved):
Looks good to me. All deprecated API has been removed and everything still works as expected!
👍🏼

PR #379, opened 2023-09-19
Fix typing of RequestActions

tortmayr:
Add an artificial property to the `RequestAction` definition to ensure that the generic type literal is not unused. This ensure that the TS compiler has all required type information and can implicitly derive the type of the Response action from a request action.

Fixes https://github.com/eclipse-sprotty/sprotty/issues/378


spoenemann (approved):
👍

PR #381, opened 2023-09-29
Renamed ViewportRootElement

spoenemann:
The following changes are for consistency with the new naming scheme:
 - Renamed `ViewportRootElement` class to `ViewportRootElementImpl`
 - Renamed `ViewportRootElementSchema` interface to `ViewportRootElement` and moved to `sprotty-protocol` package.


planger (approved):
I found one more usage in a test that hasn't been adapted and fixed in 42b2c22.
Looks great now! Thank you!

PR #382, opened 2023-10-02
Updated project dependencies

spoenemann:
In preparation of v1.0.0

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -3,26 +3,26 @@
   "version": "0.14.0",
   "private": "true",
   "dependencies": {
-    "@types/ws": "^8.5.4",
-    "@vscode/codicons": "^0.0.32",
+    "@vscode/codicons": "^0.0.33",
     "express": "^4.18.2",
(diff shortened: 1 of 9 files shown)


examples/package.json
```diff
@@ -3,26 +3,26 @@
   "version": "0.14.0",
   "private": "true",
   "dependencies": {
-    "@types/ws": "^8.5.4",
-    "@vscode/codicons": "^0.0.32",
+    "@vscode/codicons": "^0.0.33",
     "express": "^4.18.2",
+    "inversify": "^6.0.1",
```
tortmayr, line 8: This is conflicing with the `peerDependency` definition of inversify used in the sub packages (~6.0.1).
IMO there is no need to restrict this to 6.0.x so we should update the peerDependency to ~6.0.1

tortmayr (changes requested):
Looks good to me in general. Everything still works as expected.
I only noticed I minor inconsistency between the inversify peer/dependency definitions.

PR #385, opened 2023-10-04
Add request context to avoid collisions in request IDs

### LINKED ISSUE #384: Conflicting request IDs in client and server
The following code is used both by the client and the server:

https://github.com/eclipse-sprotty/sprotty/blob/08dc794c4055ed8d3b6c04b256c4a03d917a1d60/packages/sprotty-protocol/src/actions.ts#L59-L65

Problem: the generated IDs are not unique because client and server run in different JavaScript execution contexts.

Consequence: ResponseActions are sometimes associated with the wrong RequestActions.

spoenemann:
Fixes #384. Requests created in the client are always prefixed with `client_`. In addition, server code can use `setRequestContext('server')` to be 100% sure.


tortmayr (approved):
LGTM!

PR #389, opened 2023-10-06
Remove SModelExtension interface

### LINKED ISSUE #387: Remove `SModelExtension` interface
The `SModelExtension` interface in its current form serves no purpose.
I guess the original idea was to use this as marker interface for all model feature extensions.
However, the concept of empty marker interfaces is something that works well in other strongly-typed languages e.g. Java but
does not translate to typescript.
Since you cannot use an instanceof check for interfaces (and a typeguard for an empty interface is also not an option)
the interface effectively has no purpose. We should probably just remove it before the 1.0.0 release.
@spoenemann wdyt?

tortmayr:
Remove the empty `SModelExtension` interface and adapt type doc of feature extension interfaces.

Fixes #387


packages/sprotty/src/features/move/model.ts
```diff
@@ -16,15 +16,16 @@
...
 import { SModelElementImpl } from '../../base/model/smodel';
-import { SModelExtension } from '../../base/model/smodel-extension';
 
 export const moveFeature = Symbol('moveFeature');
 
 /**
  * An element that can be placed at a specific location using its position
  * property.
+ *
+ * Feature extension interface for {@link editLabelFeature}.
```
spoenemann, line 26: The feature mentioned here looks wrong.

spoenemann (approved):
Thanks!

PR #393, opened 2023-11-01
migrated from mocha to vitest

jonah-iden:
test now run via vitest. Otherwise there should have not been much change.
- there is now a `yarn coverage` command to chekc the test coverage
- tests can both be executed from root to execute all or for single packages
- there are debug targets for executing all tests and executing single test files

### DIFF
--- a/packages/sprotty/tsconfig.json
+++ b/packages/sprotty/tsconfig.json
@@ -8,5 +8,7 @@
   ],
   "exclude": [
     "**/*.spec.ts", "**/*.spec.tsx", "src/utils/test-helper.ts"
-  ]
+  ],
+  "experimentalDecorators": true,
+  "emitDecoratorMetadata": true
 }
--- a/packages/sprotty/tsconfig.json
+++ b/packages/sprotty/tsconfig.json
@@ -8,7 +8,5 @@
   ],
   "exclude": [
     "**/*.spec.ts", "**/*.spec.tsx", "src/utils/test-helper.ts"
-  ],
-  "experimentalDecorators": true,
-  "emitDecoratorMetadata": true
+  ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -2,41 +2,25 @@
     "version": "0.2.0",
     "configurations": [
         {
-            "name": "Mocha: Test Current File",
+            "name": "Vitest: Run All",
             "type": "node",
             "request": "launch",
-            "cwd": "${workspaceRoot}",
-            "program": "${workspaceRoot}/node_modules/.bin/ts-mocha",
-            "args": [
-                "${file}",
-                "--no-timeouts"
-            ],
-            "env": {
-                "TS_NODE_PROJECT": "${workspaceRoot}/tsconfig.json"
-            },
-            "sourceMaps": true,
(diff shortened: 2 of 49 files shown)


packages/sprotty/tsconfig.json
```diff
@@ -8,5 +8,7 @@
   ],
   "exclude": [
     "**/*.spec.ts", "**/*.spec.tsx", "src/utils/test-helper.ts"
-  ]
+  ],
+  "experimentalDecorators": true,
+  "emitDecoratorMetadata": true
```
spoenemann, line 13: Shouldn't these be part of the `compilerOptions`? The base tsconfig should already include them.

spoenemann (approved):
Thanks! Just one detail

packages/sprotty/tsconfig.json
jonah-iden: ups yeah, i think forgot to remove them again. I'll do that right away

dhuebner:
@jonah-iden 
I have in some spec files eslint errors: e.g. 'missing semicolon' in diagram-server.spec.ts. Are thees new because of eslint configuration changes? 
An additional question: Is it possible with vitest to somehow `import 'reflect-metadata';` by default for all tests?

jonah-iden:
@dhuebner these eslint errors seem to have been there before, they are also there on master. I didn't change anything realted to eslint.
Regarding your second question, i searched for a solution while implementing this, but sadly couldn't find anything. If you know of any i would be happy to add that

spoenemann:
Currently we exclude test files when we execute eslint:
https://github.com/eclipse-sprotty/sprotty/blob/727c2246715fc633be1e3cc65bfcdf4fce9ed359/packages/sprotty/package.json#L50

But the ESLint extension shows problem markers for the test files anyway. I propose to remove the exclusion of test files and apply eslint to the whole code base. WDYT?

dhuebner:
@jonah-iden 
Thank you for clarification! 

@spoenemann 
I think it is a good idea to generalize the eslint settings. One problem I see, is that when writing test code there are sometime situations (unused parameters, reflection) where one needs to write code that is then marked as error. That why I think it would be nice to have a separate config file for *.spec*.ts where we can overwrite some settings.  
I also noticed that when working with examples no checks are reported in the editor but the build fails then with e.g. unused imports ` TS6133: 'Action' is declared but its value is never read.`.

spoenemann:
Separate config files don't work well, the editor will always decide for one config. What we can do is add _ignore_ comments where needed to suppress the eslint errors/warnings.

kaisalmen:
@jonah-iden @dhuebner and @spoenemann most promising approach with workspaces and eslint so far I had was to you have one specific tsconfig (or shared one at workspace top with vscode) for eslint that includes all code from all packages and then have only one eslint script in the root that scans everything ([example](https://github.com/TypeFox/monaco-components/blob/main/.eslintrc.cjs))

spoenemann:
@kaisalmen @jonah-iden or @dhuebner, would one of you like to create a PR for that?

kaisalmen:
✋

kaisalmen:
See https://github.com/eclipse-sprotty/sprotty/pull/400

PR #394, opened 2023-11-01
watch all sprotty packages with a single command from root

jonah-iden:
added a new watch task to the root package json to watch all packages concurrently


spoenemann (commented):
I would prefer to switch to compiling with a TypeScript workspace like it's done in Langium:
https://github.com/eclipse-langium/langium/blob/main/tsconfig.build.json

Then building or watching the whole workspace can be done with a single `tsc -b tsconfig.build.json` call (add `-w` or `--watch` for watching). The only part where we'd still need `concurrently` is to bundle the examples with Webpack.

package.json
```diff
@@ -3,7 +3,8 @@
   "scripts": {
     "prepare": "lerna run prepare",
     "clean": "lerna run clean",
-    "build": "lerna run build",
+    "build": "tsc -b tsconfig.build.json",
```
spoenemann, line 6:
```suggestion
    "build": "tsc -b tsconfig.build.json && yarn --cwd examples build",
```

spoenemann (commented):
Thanks! We also need to add `references` to reflect the dependencies between packages. See how it's done in Langium:
https://github.com/eclipse-langium/langium/blob/9b177376c01b0e2c69cdbd82f5fc07b2e1d4577c/packages/langium-cli/tsconfig.src.json#L7-L9

sprotty has a dependency to sprotty-protocol and sprotty-elk has a dependency to both sprotty and sprotty-protocol.

spoenemann (commented):
Great! A few more things to top this off:
 - Remove packages/generator-sprotty/tsconfig.tsbuildinfo
 - Remove the "prepare" script from all packages and examples
 - Change the root "prepare" script to `yarn run clean && yarn run build`

kaisalmen:
> * Change the root "prepare" script to yarn run clean && yarn run build

@spoenemann why do we do that? `install` and `build` are different steps. Putting that in prepare always couples it. This gets in the way and a full build is not always desirable after `install`.

spoenemann:
For convenience, I guess. So far, it has never got in my way (It's been like this since 2017). But I'm ok with changing that for consistency with other project. But then we also need to adapt .gitpod.yml and the GitHub action workflows.

spoenemann (approved):

spoenemann:
@kaisalmen feel free to do that change in another PR.

PR #395, opened 2023-11-02
Better developer expereience for missing views

### LINKED ISSUE #368: Bad dev experience with edge selection and routing
Edges are selectable by default. And the default behavior when selecting an edge is to enable its editing mode, which makes explicit routing handles visible. But these are only visible if the respective element types are registered like this:
```typescript
        configureModelElement(context, 'routing-point', SRoutingHandleImpl, SRoutingHandleView);
        configureModelElement(context, 'volatile-routing-point', SRoutingHandleImpl, SRoutingHandleView);
        configureModelElement(context, 'bezier-create-routing-point', SRoutingHandleImpl, SBezierCreateHandleView);
        configureModelElement(context, 'bezier-remove-routing-point', SRoutingHandleImpl, SBezierCreateHandleView);
        configureModelElement(context, 'bezier-routing-point', SRoutingHandleImpl, SBezierControlHandleView);
```

Otherwise strange overlapping messages are shown in the diagram:
Screenshot 2023-07-17 at 13 54 43

We should find a way to improve the developer experience in this regard.

jonah-iden:
Fixes #368
stopped missing view messages from overlapping;
messages now containing more relevant information; 
added warning to console

**How to test:**
using the class diagram example,
comment out: 
```
 configureModelElement(context, 'routing-point', SRoutingHandleImpl, SRoutingHandleView);
 configureModelElement(context, 'volatile-routing-point', SRoutingHandleImpl, SRoutingHandleView);
 configureModelElement(context, 'bezier-create-routing-point', SRoutingHandleImpl, SBezierCreateHandleView);
 configureModelElement(context, 'bezier-remove-routing-point', SRoutingHandleImpl, SBezierCreateHandleView);
 configureModelElement(context, 'bezier-routing-point', SRoutingHandleImpl, SBezierControlHandleView);
```
after starting selecting an edge you should see the missing view messages not overlapping.
In the console there should be warnings about missing views

### DIFF
--- a/packages/sprotty/src/base/views/view.tsx
+++ b/packages/sprotty/src/base/views/view.tsx
@@ -107,6 +107,7 @@ export class ViewRegistry extends InstanceRegistry<IView> {
     }
 
     override missing(key: string): IView {
+        console.warn(`no registered view for type '${key}', please configure a view in the ContainerModule`);
         return new MissingView();
     }
 }
@@ -155,8 +156,20 @@ export class EmptyView implements IView {
  */
 @injectable()
 export class MissingView implements IView {
+    private static positionMap = new Map<string, Point>();
+
     render(model: Readonly<SModelElementImpl>, context: RenderingContext): VNode {
-        const position: Point = (model as any).position || Point.ORIGIN;
-        return <text class-sprotty-missing={true} x={position.x} y={position.y}>?{model.id}?</text>;
+        const position: Point = (model as any).position || this.getPostion(model.type);
+        return <text class-sprotty-missing={true} x={position.x} y={position.y}>missing "{model.type}" view</text>;
+    }
+
+    getPostion(type: string) {
+        let position = MissingView.positionMap.get(type);
+        if (!position) {
+            position = Point.ORIGIN;
+            MissingView.positionMap.forEach(value => position = value.y >= position!.y ? {x: 0, y: value.y + 20} : position);
+            MissingView.positionMap.set(type, position);
+        }
+        return position;
     }
 }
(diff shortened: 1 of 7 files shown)


packages/sprotty/src/base/views/view.tsx
```diff
@@ -155,8 +156,20 @@ export class EmptyView implements IView {
  */
 @injectable()
 export class MissingView implements IView {
+    private static positionMap = new Map<string, Point>();
+
     render(model: Readonly<SModelElementImpl>, context: RenderingContext): VNode {
-        const position: Point = (model as any).position || Point.ORIGIN;
-        return <text class-sprotty-missing={true} x={position.x} y={position.y}>?{model.id}?</text>;
+        const position: Point = (model as any).position || this.getPostion(model.type);
+        return <text class-sprotty-missing={true} x={position.x} y={position.y}>missing "{model.type}" view</text>;
```
spoenemann, line 163: Currently only a few examples include CSS for the `sprotty-missing` class. I think we should move this to the main sprotty.css file. I'd also include some text alignment property so the messages are aligned left instead of centered.

packages/sprotty/src/base/views/view.tsx
```diff
@@ -107,6 +107,7 @@ export class ViewRegistry extends InstanceRegistry<IView> {
     }
 
     override missing(key: string): IView {
+        console.warn(`no registered view for type '${key}', please configure a view in the ContainerModule`);
```
spoenemann, line 110: Use the ILogger API to log warnings (inject with `TYPES.ILogger`).

PR #396, opened 2023-11-03
fixed (w/h)Align center for containers with min width

### LINKED ISSUE #106: vbox and hbox layout do not center children properly in presence of 'minWidth', 'maxWidth' option
We created a model that contains some simple nodes that basically consists of a simple 'box' with a label inside of it.

We specify a minWidth and maxWidth on the container.

The label inside the container do not get centered even though we specify that as an option on the label.

What seems to be happening is that options to center the children are taken into account, but only considering the maximum width of the children, independent of container's minWidth/heigth. 

So basically,  let's consider a 'vbox' as example. If we place multiple children then they *are* centered with respect to eachother, and this looks fine *if* the largest child is wider than minWidth (i.e. if the children dictate the width of the container).

On the other hand, if the minWidth is larger than the width of the largest child then the children are only centered relative to eachother, but not relative to the *larger* container. And then the children end-up with the largest child stuck to the left side of the container.

Similar behavior is also observed with hbox (centering  vertically in this case).

I tried to understand the sprotty VBox layouting code a little and I think the problem is that in here:

https://github.com/eclipse/sprotty/blob/master/src/features/bounds/abstract-layout.ts#L59

the layout is adjusting the final container width taking minWidth and minHeight into account but...

- these options are only factored in at the end, not during layout of children and
- the correction 'minWidth/heigth' applied at the end don't account for alignment options

I assume this is not intentional and so to be considered a bug?

I do not have a simple example to reproduce this unfortunately. I may try to create one later.

jonah-iden:
Fixes #106

This fixes center alignment for when the parent node or the compartments parent has a set minWidth.

**How to test:**
Take the class diagram example 
configure the first node as follows (model-source.ts:155)
```
            layout: 'vbox',
            layoutOptions: {
                minWidth: 200,
            },
            children: [
                <SCompartment>{
                    id: 'node0_header',
                    type: 'comp:header',
                    layout: 'vbox',
                    layoutOptions: {
                        hAlign: 'center',
                    },
                    children: [
                        <SLabel>{
                            id: 'node0_classname',
                            type: 'label:heading',
                            text: 'Foo'
                        },
                        <SLabel>{
                            id: 'node0_classname_2',
                            type: 'label:heading',
                            text: 's'
                        },
                    ]
```
you shouold now see both text items are aligned correctly in the center. Previous to this fix they would be aligned left even though hAlign center was set

### DIFF
--- a/packages/sprotty/src/features/bounds/abstract-layout.ts
+++ b/packages/sprotty/src/features/bounds/abstract-layout.ts
@@ -32,11 +32,11 @@ export abstract class AbstractLayout<T extends AbstractLayoutOptions> implements
         const childrenSize = this.getChildrenSize(container, options, layouter);
         const maxWidth = options.paddingFactor * (
             options.resizeContainer
-            ? childrenSize.width
+            ? Math.max(childrenSize.width, options.minWidth)
             : Math.max(0, this.getFixedContainerBounds(container, options, layouter).width) - options.paddingLeft - options.paddingRight);
         const maxHeight =  options.paddingFactor * (
             options.resizeContainer
-            ? childrenSize.height
+            ? Math.max(childrenSize.height, options.minHeight)
             : Math.max(0, this.getFixedContainerBounds(container, options, layouter).height) - options.paddingTop - options.paddingBottom);
         if (maxWidth > 0 && maxHeight > 0) {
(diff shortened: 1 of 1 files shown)


spoenemann (approved):
Thanks!

PR #400, opened 2023-11-07
Eslint is executed from the root

kaisalmen:
Move eslint execution to the main `package.json`.
Remove the need for base configuration, This applies both to tsconfig and eslint config. Every sub-tsconfig defines their own includes and excludes. VSCode configures its behaviour from the global the `tsconfig.json`, too that's why it makes sense to feed both eslint and vscode from the same file.
Removed unneeded exclusions and fixed all eslint issues.
Included `lint` step in GitHub Actions.


tsconfig.json
```diff
@@ -1,6 +1,39 @@
...
+        "jsx": "react"
+    },
     "include": [
-        "packages/*/src/**/*.ts",
+        "**/src/**/*.ts",
+        "**/src/**/*.tsx",
+        "**/packages/*/*/*.ts",
+        "**/packages/*/*/*.js",
+        "*.js",
+        "*.ts"
```
spoenemann, line 33: Why so redundant?

spoenemann (approved):
Thanks!

tsconfig.json
kaisalmen: It is just for configuration files (vite and eslint).
It is actually wrong I just realized. It should be:
```
"**/packages/*/*.ts",
"**/packages/*/*.js"
```

spoenemann: What files are you including with those two entries?

kaisalmen: Any `vite.config.ts` or `.eslintrc.js`. Lint is checked there as well as well as VSCode is looking at them. All packages have their own include definition.

spoenemann: I don't think it's necessary to lint the configs.

kaisalmen: Ok, remove below. I will merge once the build pipeline was successful.

kaisalmen:
Sorry, I produced a merge commit.

PR #401, opened 2023-11-08
Deselect nodes even if select feature of clicked is disabled

### LINKED ISSUE #128: Deselect when select feature is disabled
It could be interesting to still allow de-selection when the select feature is disabled for a certain component. For example I have a node with some inner nodes where the parent node has the select feature disabled. Now in order to deselect a node I have to click outside of the parent node, as clicking the parent node does nothing to the current selection

jbicker:
Fixes #128

deselects selected nodes
even if the clicked element has no selectFeature enabled

### DIFF
--- a/packages/sprotty/src/features/select/select.ts
+++ b/packages/sprotty/src/features/select/select.ts
@@ -132,11 +132,13 @@ export class SelectMouseListener extends MouseListener {
 
     wasSelected = false;
     hasDragged = false;
+    isMouseDown = false;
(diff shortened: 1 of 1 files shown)


spoenemann (approved):
The behavior after this change looks good. But I'd like to have a second tester to be sure, @dhuebner would you have a look, too?

dhuebner (approved):
I tested with all the examples. Works as expected.

PR #399, opened 2023-11-07
Remove dependency to jsdom and update tests

gfontorbe:
Replace dependency to 'jsdom' by 'happy-dom' due to license conflicts.

Also enable tests on .tsx files.

PR #402, opened 2023-11-09
Enables jsx tests

dhuebner:
Enables jsx tests again. Also removed dev dependency to `@types/chai`.

To test run all Vitest tests:
- Tests before: 198
- Tests after: 210


dhuebner:
Already handled in #399

PR #404, opened 2023-11-10
Replace rimraf with shx

kaisalmen:
`shx` is the command-line util of `shelljs` which is already a dependency of `yeoman-generator`
I just realized `rimraf` was updated again after a long pause. Now `shelljs` is moving slow, but there still seems to be activity.

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -26,7 +26,8 @@
   },
   "scripts": {
     "prepare": "yarn run build",
-    "build": "webpack && tsc",
+    "compile": "tsc -b tsconfig.json",
+    "build": "webpack && npm run compile",
     "watch": "yarn run watch:browser",
     "watch:browser": "webpack --watch",
     "watch:server": "tsc -w",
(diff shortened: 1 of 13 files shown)


spoenemann (approved):
Thanks!

I noticed that generator-sprotty doesn't produce any output, and I don't see why. Could you have a look?

kaisalmen:
> I noticed that generator-sprotty doesn't produce any output, and I don't see why. Could you have a look?

I will

kaisalmen:
Fixed: Clean in `generator-sprotty` did not delete `*.tsbuildinfo`

I have remove package level build, watch and test scripts, too. Is this too radical? Performance wise this is not a problem. The question is if people will use something else then the top-level ones. WDYT?

spoenemann:
If we can keep the package-level scripts at no cost, I think we should do that. They may not be used frequently, but it can still be useful to have them.

kaisalmen:
Done!

dhuebner:
@kaisalmen @spoenemann 
Works well for me!

Issue #392, opened 2023-10-25
Edge created while dragging from a port is deleted after any model change


claudemamo:
Not sure if this is expected behaviour but an edge that is created while dragging from a port is deleted after any change to the model such as moving a node:

![Recording 2023-10-25 at 18 56 05](https://github.com/eclipse-sprotty/sprotty/assets/823038/74550287-4cfe-4f08-8bf8-f90ad70d2240)
 
The port is implemented as follows:

```typescript
export class StructPort extends SPortImpl implements CreatingOnDrag {
    createAction(id: string): Action {
        const edge: SEdge = {
            id,
            type: 'edge',
            sourceId: this.id,
            targetId: this.parent.id
        };
        return CreateElementAction.create(edge, { containerId: this.root.id });
    }
}
```

`MoveMouseListener` is deleting edges that have the ID `edge-in-progress` but it's also deleting edges that are no longer in-progress. As a workaround, I've written a custom router that changes the edge ID `edge-in-progress` to a random ID :

```typescript
export class CustomRouter extends PolylineEdgeRouter {
    applyReconnect(edge: SRoutableElementImpl, newSourceId?: string, newTargetId?: string) {
        let hasChanged = false;
        if (newSourceId) {
            const newSource = edge.root.index.getById(newSourceId);
            if (newSource instanceof SConnectableElementImpl) {
                edge.sourceId = newSource.id;
                hasChanged = true;
            }
        }
        if (newTargetId) {
            const newTarget = edge.root.index.getById(newTargetId);
            if (newTarget instanceof SConnectableElementImpl) {
                edge.targetId = newTarget.id;
                hasChanged = true;
            }
        }
        if (hasChanged) {
            // reset attached elements in index
            edge.index.remove(edge);
            if (edge.id === edgeInProgressID)
                edge.id = "edge-" + Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
            edge.index.add(edge);
            if (this.getSelfEdgeIndex(edge) > -1) {
                edge.routingPoints = [];
                this.cleanupRoutingPoints(edge, edge.routingPoints, true, true);
            }
        }
    }
}
```

dhuebner:
@spoenemann 
Should the `abstract-edge-router` give the new connected edge a proper id after connect in `applyReconnect`? Or how is it supposed to work with the `edgeInProgressID`?

spoenemann:
[...]

PR #405, opened 2023-11-15
ID for edge in progress after connect

dhuebner:
When connecting a new created edge inside the `applyReconnect` function a new valid and unique id will be set instead of generic `edge-in-progress`. Doing so the new edge will remain visible after an update.  

I tested the behavior with the random-graph example project. 

See #392


spoenemann (commented):
I tested on the branch you mentioned and got
```
Error: Duplicate ID in model: edge-in-progress-target-anchor
```
Sometimes I also saw an other issue where the edge-in-progress does not follow the mouse cursor, but is shifted by some amount 🤔

dhuebner:
@spoenemann 
Yes, I notice it as well. Especially if one lose the focus (e.g. because of breakpoint) and goes back to the diagram. I don't think it is related to this change, but to the move handler. We need definitely look into it. I will spend some time to investigate it.

spoenemann:
Could you look into the `edge-in-progress-target-anchor`? Maybe we have to rename or remove that, too.

dhuebner:
They are normally removed automatically after you click somewhere. I don't think we should remove them right after reconnect. But maybe re-naming them is a good idea. We just need to make sure they are removed again when the edge is deselected.

packages/sprotty/src/features/routing/abstract-edge-router.ts
```diff
@@ -297,6 +297,25 @@ export abstract class AbstractEdgeRouter implements IEdgeRouter {
...
+            if (edge.id === edgeInProgressID) {
+                // create a proper edge id after connecting the edge in progress
+                const idGen = (counter: number) => `${edge.sourceId}_to_${edge.targetId}_${counter}`;
+                let idx = 0;
+                let newId = idGen(idx);
+                while (edge.index.getById(newId) !== undefined) {
+                    newId = idGen(++idx);
+                }
+                edge.id = newId;
+                const progressTargetHandle = edge.children.find(child => child.id === edgeInProgressTargetHandleID);
```
dhuebner, line 309: @spoenemann  I've added a clean-up code for the target handle and the dangling anchor

spoenemann (approved):
👍

PR #406, opened 2023-12-04
Fix SVG export of styles

### LINKED ISSUE #149: SVG export of styles in Firefox is broken.
I am looking at a graph generated by the theia-xtext-sprotty-example project. Both in Chrome and in Firefox the graphs look good when viewed in Theia. After using the export feature (alt+E) on the diagram, only the Chrome-exported SVG works in most viewers, the Firefox-exported one does not work in most viewers, even Firefox itself cannot display the SVG anymore.

These are the SVGs exported by the different browsers:
[MrsGrantsSecretCompartment.zip](https://github.com/eclipse/sprotty/files/3896997/MrsGrantsSecretCompartment.zip)
The first file is the Chrome-generated, the one ending in '1' is generated by Firefox and the file ending with '1-modified' is the same file by Firefox with the topmost style attribute in the <svg> tag replaced with the style attribute of the Chrome-SVG.
What can be seen there is that the exporter includes way too much CSS style properties of Theia not needed in the SVG, most notably the '--theia-icon-open-json' as an entire icon in base64.

This is the statemachine used to generate the diagrams:
```
statemachine MrsGrantsSecretCompartment

event doorClosed
event drawOpened
event lightOn
event panelClosed

state active
    lightOn => waitingForDraw
    drawOpened => waitingForLight

state idle
    doorClosed => active

state waitingForLight
    lightOn => idle

state waitingForDraw
    drawOpened => unlockedPanel

state unlockedPanel
    panelClosed => idle
```

### LINKED ISSUE #407: SVG export doesn't work when ProjectedViewport is used
Because the `svg` element is wrapped in a root `div` the svg-exporter fails to find an SVG element to export.
Reproducible with the SVG example.

dhuebner:
- Filter out `inline-size`style information when copying styles from the `sprotty-hidden` elements (Fixes #149).
- More robust element look-up for 'svg' elements when exporting. (Fixes #407)

To test use some of example diagrams. Press `CMD/Ctrl+Shift+E` to export an SVG.

Updated copyright information in examples. Also added short-cut information as a footer to the class diagram example.

### DIFF
--- a/packages/sprotty/src/features/export/svg-exporter.ts
+++ b/packages/sprotty/src/features/export/svg-exporter.ts
@@ -50,12 +50,20 @@ export class SvgExporter {
 
     export(root: SModelRootImpl, request?: RequestAction<ExportSvgAction>): void {
         if (typeof document !== 'undefined') {
-            const div = document.getElementById(this.options.hiddenDiv);
-            if (div !== null && div.firstElementChild && div.firstElementChild.tagName === 'svg') {
-                const svgElement = div.firstElementChild as SVGSVGElement;
-                const svg = this.createSvg(svgElement, root);
-                this.actionDispatcher.dispatch(ExportSvgAction.create(svg, request ? request.requestId : ''));
+            const hiddenDiv = document.getElementById(this.options.hiddenDiv);
+            if (hiddenDiv === null) {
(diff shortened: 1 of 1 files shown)


spoenemann (approved):
Thanks!

PR #409, opened 2023-12-06
Smaller improvements in sprotty-protocol

spoenemann:
- Action handlers for the DiagramServer may return void.
- Added an optional index to the DiagramState (clients need to create it with their own strategy – doing it generically might be too costly).


dhuebner (approved):
LGTM

PR #410, opened 2023-12-06
Update vite and vitest, compile and fix tests

kaisalmen:
`vite` is now explicitly included as devDependency, so its version is more transparent.

There were a couple of undetected compilation errors in spec files. Therefore special `tsconfig.test.json` with `noEmit=true` have been added. They are included in the main compilation now. `tsconfig.src.json` cover the regular code compilation.  `tsconfig.json` are required for vscode/eslint only now (we do the same in other projects, e.g. Langium). Alle `"type": []` usages has been removed from **tsconfig** files as they shadows type issues. Therefore `skipLibCheck` is needed in test configs to shadow snabbdom and rollup type definition errors.

Node version in GitHub Action is upgraded to node 18. vite v5 requires it.
Volta instructions have been added.
Test were unstable with new default setting. Reason is unknown. It could be inversify not being safe with regard to concurrency, but that is speculation. pool is now configured to use only one worker thread.

`inversify` and `reflect-metadata` have been updated to `~6.0.2` and `~0.1.14` respectively.


kaisalmen:
This should have been prevented by fixing the version to `6.0.1.` (see description). It does not fail in GHA, that's why I think it is configured correctly. What does `yarn list inversify` say?

dhuebner:
It says:

```
gitpod /workspace/sprotty (kaisalmen/vite-vitest-update) $ yarn list inversify
yarn list v1.22.21
warning Filtering by arguments is deprecated. Please use the pattern option instead.
└─ inversify@6.0.1
```

dhuebner:
@kaisalmen 
I also noticed that the tasks.json is outdated. Maybe it can be fixed in this PR as well?

kaisalmen:
> Just remove "* All" and "* Example" tasks.

Done.

vscode related problems should be fixed.

Tests are instable. Happens more often in the pipeline. Don't know why, yet.

examples/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "1.0.0",
   "private": "true",
   "dependencies": {
-    "@vscode/codicons": "^0.0.33",
+    "@vscode/codicons": "~0.0.35",
     "express": "^4.18.2",
-    "inversify": "^6.0.1",
+    "inversify": "6.0.1",
```
spoenemann, line 8: I'm not sure about this change. Why not allow compatible versions? There's also 6.0.2 now.

spoenemann (commented):
Running the tests in Gitpod I get
```
 FAIL  packages/sprotty/src/base/commands/command-stack.spec.ts [ packages/sprotty/src/base/commands/command-stack.spec.ts ]
Error: Expression expected
 ❯ error node_modules/rollup/dist/es/shared/parseAst.js:337:30
 ❯ nodeConverters node_modules/rollup/dist/es/shared/parseAst.js:2084:9
 ❯ convertNode node_modules/rollup/dist/es/shared/parseAst.js:969:12
 ❯ convertProgram node_modules/rollup/dist/es/shared/parseAst.js:960:48
 ❯ parseAstAsync node_modules/rollup/dist/es/shared/parseAst.js:2150:20
 ❯ ssrTransformScript node_modules/vite/dist/node/chunks/dep-YJaePtkC.js:49505:15
 ❯ loadAndTransform node_modules/vite/dist/node/chunks/dep-YJaePtkC.js:49093:11
```

examples/package.json
kaisalmen: From the description: "The inversify version has been fixed to 6.0.1 as 6.0.2 introduces breaking changes. This became required after re-generation yarn.lock"

kaisalmen: I will revert the `yarn.lock` regeneration.

kaisalmen:
@dhuebner and @spoenemann squashed all commits into one and removed all unrelated dependency changes. Dependency changes are only related to vite and vitest. vite v5 itself require node 18, so I kept that change.

All compiler adjustments and test fixes are kept as well.
I added **volta** instructions to all packages.json now, so everyone who uses it, now always has correct node and yarn versions.

The vitest instability is still not resolved. I am digging into this now...

kaisalmen:
@dhuebner and @spoenemann this is now ready for review. Description is now up-to-date with performed changes.

.eslintrc.js
```diff
@@ -10,7 +10,8 @@ module.exports = {
     parser: '@typescript-eslint/parser',
     ignorePatterns: [
         '**/{node_modules,lib}',
-        '**/sprotty-local-template/**/*'
+        '**/sprotty-local-template/**/*',
+        '**/lib/**/src/**/*'
```
dhuebner, line 14: Does the pattern above `'**/{node_modules,lib}'` fail to ignore the lib content?

dhuebner (approved):
No inversify type errors in the IDE, nice! :-)  Tests are passing.
One thing we need to keep an eye on are regex inclusion exclusion patterns. We have a folder named `lib` in `packages/sprotty/src/lib`  and the output folders are also named `lib`. However, right now it seams to work well!

packages/sprotty-elk/typings/elkjs/index.d.ts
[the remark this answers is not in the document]
kaisalmen: elk supplies it (maybe that was different in the past) and it was producing problems as TypeScript preferred this file.

.eslintrc.js
kaisalmen: Yes, I needed to add `'**/lib/**/src/**/*'`

kaisalmen: As otherwise source folders in the new lib folder were considered.

There is room for optimization in the examples, but in another PR. 🙂

PR #411, opened 2023-12-15
Move edge labels

jbicker:
This PR introduces a new configuration option for enhanced label positioning related to edges. It adds the possibility to move labels along the edge or completely free. Or set the moveMode to none to not move a certain label at all.

The move feature must be enabled in the di config for make label placement work.
For type-safe edge placement, the corresponding element type in the sprotty model must be extended with EdgeLayoutable type.
If edgePlacement is not specified, the label is placed at default position, with the exception that labels can be freely moved.
When edgePlacement is configured and moveMode is set to 'none', labels will not be moveable.
If edgePlacement is set and moveMode is 'edge', labels are constrained to movement along the associated edge. This also serves as the default behavior when moveMode is undefined.
When edgePlacement is configured and moveMode is 'free', labels can be moved freely in any direction.

### DIFF
--- a/packages/sprotty/src/features/edge-layout/edge-layout.ts
+++ b/packages/sprotty/src/features/edge-layout/edge-layout.ts
@@ -23,44 +23,87 @@ import { setAttr } from "../../base/views/vnode-utils";
 import { SEdgeImpl } from "../../graph/sgraph";
 import { Orientation } from "../../utils/geometry";
 import { isAlignable, BoundsAware } from "../bounds/model";
-import { DEFAULT_EDGE_PLACEMENT, isEdgeLayoutable, EdgeLayoutable, EdgePlacement } from "./model";
+import { DEFAULT_EDGE_PLACEMENT, isEdgeLayoutable, EdgeLayoutable, EdgePlacement, checkEdgePlacement } from "./model";
 import { EdgeRouterRegistry } from "../routing/routing";
+import { TYPES } from "../../base/types";
+import { ILogger } from "../../utils/logging";
 
 @injectable()
 export class EdgeLayoutPostprocessor implements IVNodePostprocessor {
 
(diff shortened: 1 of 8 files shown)


packages/sprotty/src/features/edge-layout/model.ts
```diff
@@ -61,11 +60,20 @@ export class EdgePlacement extends Object {
...
     offset: number;
+
+    /**
+     * where should the label be moved when move feature is enabled.
+     * 'edge' means the label is moved along the edge, 'free' means the label is moved freely, 'none' means the label is not moved.
+     * Default is 'edge'.
+     */
+    moveMode?: 'edge' | 'free' | 'none';
+
 }
```
spoenemann, line 71: Shall we copy this interface to `sprotty-protocol` so it can be used on the server side, and mark this `@deprecated`?

packages/sprotty/src/features/edge-layout/edge-layout.ts
```diff
@@ -23,44 +23,69 @@ import { setAttr } from "../../base/views/vnode-utils";
...
+import { TYPES } from "../../base/types";
+import { ILogger } from "../../utils/logging";
 
 @injectable()
 export class EdgeLayoutPostprocessor implements IVNodePostprocessor {
 
     @inject(EdgeRouterRegistry) edgeRouterRegistry: EdgeRouterRegistry;
+    @inject(TYPES.ILogger) protected readonly logger: ILogger;
 
     decorate(vnode: VNode, element: SModelElementImpl): VNode {
```
spoenemann, line 37: Please add comments to the code in this method to make it easier to follow.

spoenemann (commented):
Thanks! Two remarks below

spoenemann (approved):
👍

PR #412, opened 2023-12-15
Add condition to logger

jbicker:
Sometimes (especially while development) it is useful to only log under certain conditions. This small enhancement enables to write such conditional logs like this instead of if blocks and console.log:
`this.logger.condition(edge.id === 'edge1').log(this, 'Edge Bounds are: ', edge.bounds);`

### DIFF
--- a/packages/sprotty/src/model-source/logging.ts
+++ b/packages/sprotty/src/model-source/logging.ts
@@ -16,7 +16,7 @@
 
 import { injectable, inject } from 'inversify';
 import { LoggingAction } from 'sprotty-protocol/lib/actions';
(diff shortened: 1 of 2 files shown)


spoenemann (commented):
I'm not sure about that – a plain old `if` block still has advantages:
```typescript
logger.condition(bla).log(expression that is expensive to compute);
```
vs.
```typescript
if (bla) {
    logger.log(expression that is expensive to compute);
}
```
In the first case the "expression that is expensive to compute" is always evaluated. In the second case. It's only evaluated when the condition is met.

I'm afraid that the proposed API can lead to bad practices.

PR #413, opened 2024-01-03
Moved feature interfaces to sprotty-protocol

spoenemann:
I moved most of the feature interfaces to the `sprotty-protocol` package so we can use them there. The old definitions are deprecated and can be removed in v2.0.0.

### DIFF
--- a/packages/sprotty-protocol/src/model.ts
+++ b/packages/sprotty-protocol/src/model.ts
@@ -45,22 +45,14 @@ export interface ViewportRootElement extends SModelRoot, Partial<Viewport>, Part
 /**
  * Root element for graph-like models.
  */
-export interface SGraph extends ViewportRootElement {
+export interface SGraph extends ViewportRootElement, Partial<LayoutableChild> {
     children: SModelElement[]
-    layoutOptions?: ModelLayoutOptions
 
     /** @deprecated Use `position` and `size` instead. */
     bounds?: Bounds
 }
 
-/**
- * Options to control the "micro layout" of a model element, i.e. the arrangement of its content
- * using simple algorithms such as horizontal or vertical box layout.
- */
-export type ModelLayoutOptions = { [key: string]: string | number | boolean };
-
-export interface SShapeElement extends SModelElement, Partial<BoundsAware> {
-    layoutOptions?: ModelLayoutOptions
+export interface SShapeElement extends SModelElement, Partial<LayoutableChild> {
 }
 
 /**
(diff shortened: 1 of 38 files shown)


packages/sprotty-protocol/src/model.ts
gfontorbe, line 1: For code organization, I am wondering if it would make more sense to move the features interfaces into their own `features.ts` instead of `model.ts`

jbicker (approved):
LGTM so far. What Guillaume mentioned [here](https://github.com/eclipse-sprotty/sprotty/pull/413/files#r1440230423) is worth a thought though.

packages/sprotty-protocol/src/model.ts
spoenemann: Moving the existing features like `BoundsAware` would be a breaking change because with CommonJS imports you can import directly from `sprotty-protocol/lib/model`. We can consider a reorganization when we switch to ESM and package exports.

spoenemann:
@gfontorbe I moved Locateable.

The problem with LayoutableChild and LayoutContainer is that they extend BoundsAware, but the definition of BoundsAware in the external model (`position` & `size`) differs from the one in the internal model (`bounds`). I renamed the three interfaces to InternalBoundsAware, InternalLayoutableChild and InternalLayoutContainer to make the difference explicit.

The other features are either empty interfaces (Decoration, Deletable) or are used only in the frontend AFAICT.

gfontorbe (approved):

packages/sprotty-protocol/src/model.ts
```diff
@@ -45,31 +45,22 @@ export interface ViewportRootElement extends SModelRoot, Partial<Viewport>, Part
...
 
-/**
- * Options to control the "micro layout" of a model element, i.e. the arrangement of its content
- * using simple algorithms such as horizontal or vertical box layout.
- */
-export type ModelLayoutOptions = { [key: string]: string | number | boolean };
-
-export interface SShapeElement extends SModelElement, Partial<BoundsAware> {
-    layoutOptions?: ModelLayoutOptions
+export interface SShapeElement extends SModelElement, Partial<LayoutableChild> {
```
sailingKieler, line 55: `SShapeElement` is not `BoundsAware` anymore. Was that intended?
`SNode`, `SPort`, and `SLabel` don't have `position` and `size` anymore, or do I overlook something?

sailingKieler: Ah, it is inherited via `LayoutableChild`? The name is a bit misleading, isn't it?

spoenemann: Do you have an alternative proposal?

PR #414, opened 2024-01-04
Make command executions (e.g. animations) stoppable

jbicker:
Commands that implement the new introduced IStoppableCommand do stop the last/current execution immediately to start another one of the same kind.

Solves https://github.com/eclipse-sprotty/sprotty/issues/1

It is nicely testable by raising the [duration time](https://github.com/jbicker/sprotty/blob/7d028502092be1533e60410cf824d83da8bcd4c7/packages/sprotty/src/base/di.config.ts#L86) of the CommandStackOptions and trigger the scrambling in the circle example multiple times.

### DIFF
--- a/examples/circlegraph/src/standalone.ts
+++ b/examples/circlegraph/src/standalone.ts
@@ -77,11 +77,15 @@ export default async function runCircleGraph() {
         graph.children.push(...newElements);
     }
 
+    let viewport = initialViewport;
+    window.addEventListener('resize', async () => {
+        viewport = await modelSource.getViewport();
+    });
+
     // Run
     modelSource.setModel(graph);
 
     async function createNode(point?: Point) {
-        const viewport = await modelSource.getViewport();
         const newElements = addNode(getVisibleBounds(viewport));
         if (point) {
             const adjust = (offset: number) => {
@@ -103,7 +107,6 @@ export default async function runCircleGraph() {
     });
 
     document.getElementById('scrambleAll')!.addEventListener('click', async () => {
-        const viewport = await modelSource.getViewport();
         const bounds = getVisibleBounds(viewport);
         const nodeMoves: ElementMove[] = [];
         graph.children.forEach(shape => {
@@ -117,13 +120,12 @@ export default async function runCircleGraph() {
                 });
             }
         });
-        dispatcher.dispatch(MoveAction.create(nodeMoves, { animate: true }));
+        dispatcher.dispatch(MoveAction.create(nodeMoves, { animate: true, stoppable: true }));
         focusGraph();
     });
 
     document.getElementById('scrambleSelection')!.addEventListener('click', async () => {
         const selection = await modelSource.getSelection();
-        const viewport = await modelSource.getViewport();
         const bounds = getVisibleBounds(viewport);
         const nodeMoves: ElementMove[] = [];
         selection.forEach(shape => {
@@ -137,7 +139,7 @@ export default async function runCircleGraph() {
                 });
             }
         });
-        dispatcher.dispatch(MoveAction.create(nodeMoves, { animate: true }));
+        dispatcher.dispatch(MoveAction.create(nodeMoves, { animate: true, stoppable: true }));
         focusGraph();
     });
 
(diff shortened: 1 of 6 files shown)


kaisalmen (approved):
Tested successfully. Changes look good. LGTM.

examples/circlegraph/src/standalone.ts
```diff
@@ -77,11 +77,15 @@ export default async function runCircleGraph() {
...
+    let viewport = initialViewport;
+    window.addEventListener('resize', async () => {
+        viewport = await modelSource.getViewport();
+    });
+
     // Run
     modelSource.setModel(graph);
 
     async function createNode(point?: Point) {
-        const viewport = await modelSource.getViewport();
```
spoenemann, line 84: The example now starts with all nodes in the top left corner:
Screenshot 2024-02-16 at 11 13 14

packages/sprotty/src/base/commands/command.ts
```diff
@@ -52,6 +52,17 @@ export interface ICommand {
...
 
+/**
+ * A stoppable commands execution (e.g. one that starts an animation) can be interrupted.
+ */
+export interface IStoppableCommand extends ICommand {
+    stopExecution(): void
+}
+
+export function isStoppableCommand(command: any): command is IStoppableCommand {
+    return command && 'stopExecution' in command && typeof command.stopExecution === 'function';
```
spoenemann, line 63: You can use the utility function `hasOwnProperty` from sprotty-protocol here.

packages/sprotty/src/base/commands/command-stack.ts
```diff
@@ -210,6 +215,16 @@ export class CommandStack implements ICommandStack {
     protected handleCommand(command: ICommand,
                             operation: (context: CommandExecutionContext) => CommandReturn,
                             beforeResolve: (command: ICommand, context: CommandExecutionContext) => void) {
+        // If the command implements the IStoppableCommand interface, we first need to stop the execution of the
+        // previous command with the same action kind and then store the new command as the last stoppable command.
+        if (isStoppableCommand(command)) {
+            const stoppableCommand = this.stoppableCommands.get((command as any).action.kind);
```
spoenemann, line 221: This is quite dirty: `action` is not defined in the `ICommand` interface, and the property is `protected` in most cases.

Please extract the map key computation to a protected method:
```typescript
protected getStoppableCommandKey(command: IStoppableCommand): string | undefined {
    let key = Object.getPrototypeOf(command).KIND;
    if (key) {
        return key;
    }
    return (command as any).action?.kind;
}
```
If the result is `undefined`, we should skip the code here.

spoenemann (commented):
Nice solution! I have a few more remarks though. @jbicker please address them in a follow-up PR.

packages/sprotty/src/base/commands/command-stack.ts
spoenemann: Better alternative: extend the IStoppableCommand interface by a public property `keyForStopping` or similar, then the command implementation can decide how to create the key. MoveCommand can just set `keyForStopping = action.kind` in the constructor.

packages/sprotty-protocol/src/actions.ts
```diff
@@ -676,16 +676,18 @@ export interface MoveAction extends Action {
     moves: ElementMove[]
     animate: boolean
     finished: boolean
+    stoppable: boolean
```
sailingKieler, line 679: Remark: When you do such protocol changes in futures, keep in mind that there's a Java (Xtend) based implementation, too, see here: https://github.com/eclipse-sprotty/sprotty-server/blob/master/org.eclipse.sprotty/src/main/java/org/eclipse/sprotty/Actions.xtend
I've updated it for now.

packages/sprotty/src/base/commands/command.ts
jbicker: That does not work since the stopExecution is a member of the command objects prototype (MergeableCommand) and not a direct member at runtime.

examples/circlegraph/src/standalone.ts
jbicker: It starts like that since https://github.com/eclipse-sprotty/sprotty/commit/b0e9d60e9c9ab7d302f84f55eef227828ba7d68c

PR #416, opened 2024-01-19
Adds flowchart example

gfontorbe:
Adds an example for a flowchart diagram.

The main point of the example is to demonstrate custom views and in-place label editing.

Later on, this example could serve as a base to provide custom views for flowchart diagrams as a separate package.

Screenshot 2024-01-19 at 09 58 32

### DIFF
--- a/examples/flowchart/css/page.css
+++ b/examples/flowchart/css/page.css
@@ -43,7 +43,7 @@
 svg {
     margin-top: 15px;
     width: 100%;
-    height: 1000px;
+    height: 600px;
     border-style: solid;
     border-width: 1px;
     border-color: #bbb;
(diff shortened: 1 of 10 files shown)


examples/flowchart/css/page.css
```diff
@@ -0,0 +1,50 @@
...
+    text-align: right;
+    font-size: 16px;
+    color: #888;
+}
+
+svg {
+    margin-top: 15px;
+    width: 100%;
+    height: 1000px;
+    border-style: solid;
```
dhuebner, line 47: Could we change height to 600px? It still looks good, but I don't need to scroll when running the example on a smaller display. 
```suggestion
    height: 600px;
```

dhuebner (approved):
Looks good to me. Nice new example!

PR #424, opened 2024-02-07
Allow override of model element registration in DI

gfontorbe:
Allow the element registration from the dependency injection to be overridden by another `ContainerModule`. This comes in handy when working with exported containers from the `sprotty-library` or when sharing  a `ContainerModule` when only a subset of model elements need a new registration.

This introduces a new helper function `overrideModelElement` to be used instead of `configureModelElement` in the `ContainerModule`

Example:
```
const moduleA = new ContainerModule((bind, unbind, isBound, rebind) => {
...
configureModelElement(context, 'myElementType', MyElementImplA, MyElementViewA);
}

const moduleB = new ContainerModule((bind, unbind, isBound, rebind) => {
...
overrideModelElement(context, 'myElementType', MyElementImplB, MyElementViewB);
}

const container = new Container();
loadDefaultModules(container);
container.load(moduleA);
container.load(moduleB);
```
In this scenario, `myElementType` elements will first be linked to `MyElementImplA` and `MyElementViewA` before being linked to `MyElementImplB` and `MyElementViewB`. 

If the order of the functions `container.load(...)` are reversed, so is the link between the element type and its implementation/view.

Error handling:
By design, we can only override an element that has already been configured.
* Trying to use `configureModelElement` on an element type that is already registered will throw an error
* Trying to use `overrideModelElement` on an element type that is not already registered will throw an error

### DIFF
--- a/packages/sprotty/src/base/model/smodel-factory.ts
+++ b/packages/sprotty/src/base/model/smodel-factory.ts
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2017-2018 TypeFox and others.
+ * Copyright (c) 2017-2024 TypeFox and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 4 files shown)


sailingKieler:
I like it! 👍

spoenemann (approved):
Great enhancement!

Issue #296, opened 2022-05-25
Consider using `peerDepdencies` for inversify


tortmayr:
see https://github.com/eclipse/sprotty/pull/292#issuecomment-1137159348

spoenemann:
Done with #357

spoenemann:
We decided to revert the change because it is problematic in some circumstances, see https://github.com/eclipse-sprotty/sprotty/pull/429#discussion_r1492091716.

@tortmayr feel free to reopen this issue if you want to continue the discussion.

PR #429, opened 2024-02-16
Limit snabbdom to bugfix releases, remove inversify from dependencies

### LINKED ISSUE #418: Fix versions of Sprotty dependencies
When using Sprotty as part of a prototype we had issues with latest snabbdom release.
Because Sprotty just requires anything equal or larger than 3.5.1 ("snabbdom": "^3.5.1") the latest release has been resolved.

The new version of snabbdom is build on ECMA Script module system and requires updates on our side.
Please fix the version of dependencies to a concrete version or at least to just allow bug fixes only.

spoenemann:
- Exclude snabbdom 3.6.x since it causes problems for some users (closes #418)
- Revert a change made in #410 where inversify appears twice in package.json: `peerDependencies` and `dependencies`

### DIFF
--- a/packages/sprotty/package.json
+++ b/packages/sprotty/package.json
@@ -21,14 +21,11 @@
     "url": "https://github.com/eclipse-sprotty/sprotty",
     "directory": "packages/sprotty"
   },
-  "peerDependencies": {
-    "inversify": "~6.0.2"
-  },
   "dependencies": {
-    "autocompleter": "^9.1.0",
+    "autocompleter": "^9.1.2",
     "file-saver": "^2.0.5",
     "inversify": "~6.0.2",
-    "snabbdom": "^3.5.1",
+    "snabbdom": "~3.5.1",
     "sprotty-protocol": "^1.1.0",
     "tinyqueue": "^2.0.3"
   },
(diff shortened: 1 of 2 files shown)


spoenemann:
Once approved, I'll publish Sprotty 1.1.1 with this fix. Or should we apply only the snabbdom change, but exclude the inversify change in that bugfix release?

packages/sprotty/package.json
```diff
@@ -25,10 +25,9 @@
     "inversify": "~6.0.2"
   },
   "dependencies": {
-    "autocompleter": "^9.1.0",
+    "autocompleter": "^9.1.2",
     "file-saver": "^2.0.5",
-    "inversify": "~6.0.2",
```
kaisalmen, line 30: This is done intentionally, because yarn does not automatically install `peerDependencies`, I guess `inversify` is specified elsewhere, too right?

kaisalmen: npm and pnpm are ok with that when using a lib. yarn is not!

kaisalmen: @spoenemann when you remove the lock file (you get the warning). And once you remove other references to ìnversify` in the workspace (optional in elk or dependency in examples) the dependency is gone ⚠️ :
image

kaisalmen: `peerDependency` specifies the compatible version range. Important when using sprotty in another project where a conflicting version is in place.

With the change you suggest using sprotty elsewhere, npm an pnpm will fulfill a peerDependency if not met by other dependencies, yarn requires you to specify it. Therefore, specifying it here removes the problem for end-users. yarn behaving differently would be the better option, though.

kaisalmen (commented):
We need to discuss how to proceed.

packages/sprotty/package.json
spoenemann: But then I would remove the `peerDependency` and just keep it as a `dependency`, reverting the change made in #357 (as proposed in #296).

kaisalmen: Let's do this. 👍

kaisalmen (approved):
LGTM

PR #430, opened 2024-02-20
Improved JSX factory

jbicker:
Fixes https://github.com/eclipse-sprotty/sprotty/issues/419

It is now possible to add Function Components to the views as well as normal svg attributes to svg elements without adding 'attrs-' prefix.

A very simple example is added to demonstrate all cases.


spoenemann (commented):
Great, thanks! Could you add a few test cases to `jsx.spec.tsx`?

spoenemann (approved):

Issue #420, opened 2024-01-31
Junction point edge routing


glatuske:
As part of our graph we want to visualize junction points of edges (where edges split or merge) custom figure (e.g. circle).
The calculation whether a junction point is required or not should be handled on the client side by checking the routing points and same source/target.
A generic solution would be helpful for us, but could be also used by other projects (similar to other edge routings).

Issue #427, opened 2024-02-10
sprotty-elk sometimes doesn't delete old edge bends on layout changes


semira222:
The video shows the issue. When expanding/collapsing nodes, edge bends are not deleted from the layout, which can turn into quite the spaghetti on larger graphs. This only happens when the edge wouldn't need bends in the new layout. If the new edge layout just needs different bend points the old bend points are being deleted as expected.

https://github.com/eclipse-sprotty/sprotty/assets/77962181/a68bfd13-770a-4174-ae82-829d570015cb

spoenemann:
We haven't seen this in other applications using the collapse / expand feature. I think solving the issue is not possible without seeing the source code of your implementation.

semira222:
I am using elk for layouting, probably should have made that clearer. I managed to reproduce the bug in the randomgraph example from the sprotty repo by reducing the number of nodes and edges to 5 each:

https://github.com/eclipse-sprotty/sprotty/assets/77962181/5d5a1620-328a-41d7-8376-fde214ece0a3

jbicker:
@spoenemann, I can have a look a it.

spoenemann:
It turned out this is an ELK bug: https://github.com/eclipse/elk/issues/1001

PR #434, opened 2024-02-21
Implement Junction Finder

gfontorbe:
Solves #420 

Adds a new feature to find junction points on edges with the same source or target.

Similar to the `IntersectionFinder`, we provide a `ContainerModule` that can be loaded when needed in the inversify container.

The `JunctionFinder` ultimately sets the `isJunction` flag to `true` on `RoutingPoint`s where  a junction point must be drawn.

The rendering is done in the `PolylineEdgeView` (and therefore views that inherit this view) in a new function `renderFunctionPoints`.

They are represented as circles with a class `sprotty-junction` for easy styling.

Screenshot 2024-02-21 at 10 28 01

### DIFF
--- a/packages/sprotty/src/features/edge-junction/junction-finder.ts
+++ b/packages/sprotty/src/features/edge-junction/junction-finder.ts
@@ -1,0 +1,229 @@
+/********************************************************************************
+ * Copyright (c) 2021-2022 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
+ * with the GNU Classpath Exception which is available at
+ * https://www.gnu.org/software/classpath/license.html.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
+ ********************************************************************************/
+
+import { injectable } from "inversify";
+import { SParentElementImpl } from "../../base/model/smodel";
+import { SEdgeImpl } from "../../graph/sgraph";
+import { EdgeRouting, IEdgeRoutePostprocessor, RoutedPoint } from "../routing/routing";
+
+/**
+ * Finds junction points in the edge routes. A junction point is a point where two or more edges split.
+ * This excludes the source and target points of the edges.
+ *
+ * Only works with straight line segments.
+ */
+@injectable()
+export class JunctionFinder implements IEdgeRoutePostprocessor {
+    apply(routing: EdgeRouting, parent: SParentElementImpl): void {
+        this.findJunctions(routing, parent);
+    }
+
+    findJunctions(routing: EdgeRouting, parent: SParentElementImpl) {
+        // gather all edges from the parent
+        const edges = parent.children.filter(child => child instanceof SEdgeImpl) as SEdgeImpl[];
+
+        routing.routes.forEach((route, routeId) => {
+            // for each route we find the corresponding edge from the model by matching the route id and the edge id
+            const edge = edges.find(e => e.id === routeId);
+
+            if (!edge) {
+                return;
+            }
+
+            // we find all edges with the same source as the current edge, excluding the current edge
+            const edgesWithSameSource = edges.filter(e => e.sourceId === edge.sourceId && e.id !== edge.id);
+            // for each edge with the same source we find the corresponding route from the routing
+            const routesWithSameSource: RoutedPoint[][] = [];
+            edgesWithSameSource.forEach(e => {
+                const foundRoute = routing.get(e.id);
+                if (!foundRoute) {
+                    return;
+                }
+                routesWithSameSource.push(foundRoute);
+            });
+
+            // if there are any routes with the same source, we find the junction points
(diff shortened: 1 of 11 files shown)


packages/sprotty/src/features/edge-junction/junction-finder.ts
```diff
@@ -0,0 +1,229 @@
...
+ */
+@injectable()
+export class JunctionFinder implements IEdgeRoutePostprocessor {
+    apply(routing: EdgeRouting, parent: SParentElementImpl): void {
+        this.findJunctions(routing, parent);
+    }
+
+    findJunctions(routing: EdgeRouting, parent: SParentElementImpl) {
+        // gather all edges from the parent
+        const edges = parent.children.filter(child => child instanceof SEdgeImpl) as SEdgeImpl[];
```
spoenemann, line 36: If I remember correctly, the routing includes all nested subgraphs in the case of multiple hierarchy layers. For such cases, you can use `parent.index.all()`.

packages/sprotty/src/features/edge-junction/junction-finder.ts
```diff
@@ -0,0 +1,229 @@
...
+        routing.routes.forEach((route, routeId) => {
+            // for each route we find the corresponding edge from the model by matching the route id and the edge id
+            const edge = edges.find(e => e.id === routeId);
+
+            if (!edge) {
+                return;
+            }
+
+            // we find all edges with the same source as the current edge, excluding the current edge
+            const edgesWithSameSource = edges.filter(e => e.sourceId === edge.sourceId && e.id !== edge.id);
```
spoenemann, line 47: The number of routes should be more or less the same as the number of edges. This means we have quadratic runtime complexity here: with 1.000 edges, you need 1.000.000 comparisons. One way to reduce the complexity is to create multiple Maps (by `id`, `sourceId` and `targetId`) instead of a single array. However, this would likely degrade the performance for smaller graph. An optimized implementation might choose an algorithm depending on the number of routes, but I wouldn't go that far for now.

packages/sprotty/src/features/routing/routing.ts
```diff
@@ -147,7 +148,7 @@ function isMultipleEdgesRouter(
 
 /** A postprocessor that is applied to all routes, once they are computed. */
 export interface IEdgeRoutePostprocessor {
-    apply(routing: EdgeRouting): void;
+    apply(routing: EdgeRouting, parent?: SParentElementImpl): void;
```
spoenemann, line 151: I think the new argument can be mandatory. The change would be breaking only for users who are calling this service in their own code, but that's very unlikely.

spoenemann (commented):
Thanks!

In which example have you tested the new feature? Maybe we can extend the "Random Graph" example so that some edges are randomly assigned to the same port?

gfontorbe:
> In which example have you tested the new feature? Maybe we can extend the "Random Graph" example so that some edges are randomly assigned to the same port?

I modified the random graph example to fit my needs, I could probably find time to work on modifying the example to include optional junction points

packages/sprotty/src/features/edge-junction/junction-postprocessor.ts
```diff
@@ -0,0 +1,40 @@
...
+ ********************************************************************************/
+
+import { injectable } from "inversify";
+import { IVNodePostprocessor } from "../../base/views/vnode-postprocessor";
+import { VNode } from "snabbdom";
+import { Action } from "sprotty-protocol";
+import { SModelElementImpl } from "../../base/model/smodel";
+
+@injectable()
+export class JunctionPostProcessor implements IVNodePostprocessor {
```
spoenemann, line 24: Please add a comment explaining what this does.

packages/sprotty/src/features/edge-junction/junction-postprocessor.ts
```diff
@@ -0,0 +1,40 @@
...
+    decorate(vnode: VNode, element: SModelElementImpl): VNode {
+        return vnode;
+    }
+    postUpdate(cause?: Action | undefined): void {
+        const svg = document.querySelector('svg#sprotty_root > g');
```
spoenemann, line 29: Can we guarantee that the root element always has the id `sprotty_root`? Is it hard-coded somewhere in the framework?

packages/sprotty/src/features/edge-junction/junction-finder.ts
```diff
@@ -0,0 +1,241 @@
...
+import { EdgeRouting, IEdgeRoutePostprocessor, RoutedPoint } from "../routing/routing";
+
+/**
+ * Finds junction points in the edge routes. A junction point is a point where two or more edges split.
+ * This excludes the source and target points of the edges.
+ *
+ * Only works with straight line segments.
+ */
+@injectable()
+export class JunctionFinder implements IEdgeRoutePostprocessor {
```
spoenemann, line 29: This service contains complex logic that should be unit-tested. Please add a test file that covers some cases.

packages/sprotty/src/features/edge-junction/junction-postprocessor.ts
gfontorbe: No, we can't. The selector should start from the `baseDiv` specified in the viewer options.

gfontorbe:
> I'm seeing strange spider webs when I enable port merging:
> 
> Screenshot 2024-02-26 at 15 22 34
> Do you know what's going on? Is it an issue in the example or in ELK?

Thats the same issue that has been reported in #427.

Bend points are not removed so you end up with funky edges. Usually happens when you show/hide junction points several times without reloading the page. First render seems to work fine most of the time on my end.

spoenemann (approved):
👍

PR #435, opened 2024-02-21
Add supports for shadow DOM

### LINKED ISSUE #425: Support shadow dom
For now, it's not possible to tie a shadow dom element to the sprotty `baseDiv`.

Seems linked to https://github.com/eclipse-sprotty/sprotty/blob/2cdca536570640a07c1b1fb710ff8827eded4269/packages/sprotty/src/base/views/viewer.tsx#L154

gfontorbe:
Closes #425 

Adds an option in the `ViewerOptions` to add the id of the element that holds the shadow root.

The viewer will first check if this option is set before trying to retrieve the div container for the diagram.

### DIFF
--- a/packages/sprotty/src/base/views/viewer-options.ts
+++ b/packages/sprotty/src/base/views/viewer-options.ts
@@ -28,9 +28,9 @@ export interface ViewerOptions {
     hiddenDiv: string
     /** CSS class added to the base element of the hidden rendering. */
     hiddenClass: string
-    /** ID of the HTML element into which hover popup boxes are rendered. */
-    shadowRoot?: string
     /** ID of the HTML element holding the shadow root. */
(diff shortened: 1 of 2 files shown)


packages/sprotty/src/base/views/viewer-options.ts
```diff
@@ -29,6 +29,8 @@ export interface ViewerOptions {
     /** CSS class added to the base element of the hidden rendering. */
     hiddenClass: string
     /** ID of the HTML element into which hover popup boxes are rendered. */
+    shadowRoot?: string
+    /** ID of the HTML element holding the shadow root. */
```
spoenemann, line 33: The comment should be above the corresponding item.

spoenemann (approved):
Thanks!

PR #436, opened 2024-02-21
Do not set bend points on elk edge when transforming from Sprotty edge

jbicker:
Fixes https://github.com/eclipse-sprotty/sprotty/issues/427

The bend points are now completely neglected in the edge transformation from SEdge to an ElkExtendedEdge. So elk sets completely new bend points.

Feels a bit like a naive fix but it works at least for the random graph example. I hope it does not break something else.

### DIFF
--- a/packages/sprotty-elk/src/elk-layout.ts
+++ b/packages/sprotty-elk/src/elk-layout.ts
@@ -312,6 +312,10 @@ export class ElkLayoutEngine implements IModelLayoutEngine {
                 points.push(elkEdge.targetPoint);
             }
         }
+        // if x or y of first and last point are equal then use only them and leave the middle points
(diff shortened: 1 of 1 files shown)


spoenemann (commented):
In some cases, an ELK layout algorithm might use the previous bend points in its computations, so we shouldn't remove them. Could you try to fix it in the other direction, making sure that bend points are erased in the SModel when ELK doesn't specify any?

jbicker:
@spoenemann, thanks for the hint. I did as you proposed.

jbicker:
Could this PR then be closed?

spoenemann:
I think so yes. The current solution smells like a workaround for a problem that lies somewhere else.

PR #437, opened 2024-02-21
Fix zoom jump bug

### LINKED ISSUE #398: Zoom in or out results in a bad scroll offset for the last handled WheelEvent
The bug can be observed in the examples here: https://sprotty.org/
* Center the mouse in one of the Example canvas 
* Scroll in (or out)
* On the last tick before the zoom limit is reached, the diagram scrolls in a not nice way 

I think the issue is caused by a combination of ZoomMouseListener + SetViewportCommand. The ZoomMouseListener will calculate a new "zoom" which is also used to calculate the scroll values. The zoom might be smaller or bigger than the limits. The SetViewportCommand will then enforce the limits in the execute method to the exact min/max values, but leave the scroll values untouched (at least in relation to the zoom).

For now, i registered a custom ZoomMouseListener with the following processZoom method:

```
 protected processZoom(viewport: Viewport, target: SModelElementImpl, event: WheelEvent): Viewport | undefined {
    const zoomFactor = this.getZoomFactor(event);

    //Zoom min/max is reached no need to do anything
    if (
      (zoomFactor > 1 && almostEquals(viewport.zoom, this.viewerOptions.zoomLimits.max)) ||
      (zoomFactor < 1 && almostEquals(viewport.zoom, this.viewerOptions.zoomLimits.min))
    ) {
      return;
    }

    //'limitViewport' used by 'SetViewportCommand' will set the zoom to the min max level
    // so we need to do this here to to avoid 'jumps' of the diagram based on the viewport.scroll
    const zoom = limit(viewport.zoom * zoomFactor, this.viewerOptions.zoomLimits);

    const viewportOffset = this.getViewportOffset(target.root, event);
    const offsetFactor = 1.0 / zoom - 1.0 / viewport.zoom;
    const viewPort = {
      scroll: {
        x: viewport.scroll.x - offsetFactor * viewportOffset.x,
        y: viewport.scroll.y - offsetFactor * viewportOffset.y,
      },
      zoom,
    };

    return viewPort;
  }
```

gfontorbe:
Using the suggested code from @BjBe82 , fixes #398

### DIFF
--- a/packages/sprotty/src/features/viewport/zoom.ts
+++ b/packages/sprotty/src/features/viewport/zoom.ts
@@ -25,6 +25,7 @@ import { MouseListener } from '../../base/views/mouse-tool';
 import { ViewerOptions } from '../../base/views/viewer-options';
 import { getWindowScroll } from '../../utils/browser';
 import { isViewport } from './model';
+import { limit } from '../../utils/geometry';
 
 export function getZoom(label: SModelElementImpl) {
     let zoom = 1;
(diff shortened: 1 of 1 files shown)


spoenemann (approved):
👍

PR #439, opened 2024-02-22
svg export via cli command and headless browser

jbicker:
This is a first try regarding https://github.com/eclipse-sprotty/sprotty/issues/421.

Solution for now: listening to an custom browser event in the respective app (here it is the JSXample page) which will be fired by an cli command (`sprotty export http://localhost:8080/jsxample/jsxample.html`) .
Very pragmatic. Maybe there is a more general/elegant way that is to be discussed.
Hence the Draft PR.

### DIFF
--- a/packages/sprotty/package.json
+++ b/packages/sprotty/package.json
@@ -23,8 +23,10 @@
   },
   "dependencies": {
     "autocompleter": "^9.1.2",
+    "commander": "^12.0.0",
     "file-saver": "^2.0.5",
     "inversify": "~6.0.2",
+    "puppeteer": "^22.2.0",
     "snabbdom": "~3.5.1",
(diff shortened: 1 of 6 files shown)


packages/sprotty/package.json
```diff
@@ -23,8 +23,10 @@
   },
   "dependencies": {
     "autocompleter": "^9.1.2",
+    "commander": "^12.0.0",
     "file-saver": "^2.0.5",
     "inversify": "~6.0.2",
+    "puppeteer": "^22.2.0",
```
spoenemann, line 29: We should not have these dependencies in the base package of Sprotty. If we go for this solution, the new code should be moved to a new package.

spoenemann:
This is now solved with #454.

PR #442, opened 2024-03-01
Move also edge bendpoints of nested nodes when parent is moved 

### LINKED ISSUE #428: Moving multiple containers does not show feedback move of internal edge bendpoints
This seems to occur when the edges are not connected directly to the moved elements, but rather to descendants of the moved elements.

Solving this problem probably requires two fixes:

In MoveCommand.execute, only edges attached to directly moved elements are found by index.getAttachedElements(element). Perhaps the definition of outgoing / incoming edges could be extended to include edges connected to descendants, but I haven't checked for side effects to that, or an additional index data structure is needed.

Later, in MoveCommand.doMove, there is a test whether both edge ends are included in the move:
 ```
               if (edge.source
                    && edge.target
                    && this.resolvedMoves.get(edge.source.id)
                    && this.resolvedMoves.get(edge.target.id))
```
That would require checking the descendant relation as well.

### PR DESCRIPTION
**Pull request by jbicker**
Fixes #428

The classdiagram example has been changed as well in order to demonstrate this case better.

### DIFF
--- a/packages/sprotty/src/features/move/move.ts
+++ b/packages/sprotty/src/features/move/move.ts
@@ -21,7 +21,7 @@ import { Bounds, Point } from 'sprotty-protocol/lib/utils/geometry';
 import { Action, DeleteElementAction, ReconnectAction, SelectAction, SelectAllAction, MoveAction } from 'sprotty-protocol/lib/actions';
 import { Animation, CompoundAnimation } from '../../base/animations/animation';
 import { CommandExecutionContext, ICommand, MergeableCommand, CommandReturn, IStoppableCommand } from '../../base/commands/command';
-import { SChildElementImpl, SModelElementImpl, SModelRootImpl } from '../../base/model/smodel';
+import { SChildElementImpl, SModelElementImpl, SModelRootImpl, isParent } from '../../base/model/smodel';
 import { findParentByFeature, translatePoint } from '../../base/model/smodel-utils';
 import { TYPES } from '../../base/types';
 import { MouseListener } from '../../base/views/mouse-tool';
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by gfontorbe — APPROVED** [review]
Tested locally and works as expected

File: packages/sprotty/src/features/move/move.ts [hunk 1]
```diff
@@ -189,6 +202,37 @@ export class MoveCommand extends MergeableCommand implements IStoppableCommand {
         });
     }
 
+    // tests if the edge is attached to the moved element directly or to on of their children
+    protected isAttachedEdge(edge: SRoutableElementImpl): boolean {
+        const source = edge.source;
+        const target = edge.target;
+        const checkMovedElementsAndChildren = (sourceOrTarget: SConnectableElementImpl): boolean => {
+            return Array.from(this.resolvedMoves.values()).some(res => {
```
spoenemann, line 210: This check potentially traverses the whole subgraph of each moved element, which can be very large in some applications. It's much more efficient to do the traversal the other way around: from the edge source or target element, going up the parent chain, until you find an element contained in `resolvedMoves` (or not).

packages/sprotty/src/features/move/move.ts
```diff
@@ -189,6 +202,37 @@ export class MoveCommand extends MergeableCommand implements IStoppableCommand {
...
+                            return false;
+                        });
+                    }
+                    return false;
+                };
+                return recursiveCheck(res.element);
+            }) || !!(this.resolvedMoves.get(sourceOrTarget.id));
+        };
+
+        return !!(
```
spoenemann, line 228: I prefer the more explicit `Boolean(...)`

packages/sprotty/src/features/move/move.ts
```diff
@@ -106,16 +106,32 @@ export class MoveCommand extends MergeableCommand implements IStoppableCommand {
...
+                                    const existingDelta = attachedEdgeShifts.get(edge);
+                                    const newDelta = Point.subtract(resolvedMove.toPosition, resolvedMove.fromPosition);
+                                    const delta = (existingDelta)
+                                        ? Point.linear(existingDelta, newDelta, 0.5)
+                                        : newDelta;
+                                    attachedEdgeShifts.set(edge, delta);
+                                }
+                            });
+                        };
+                        const handleEdgesForChildren = (el: SModelElementImpl) => {
```
spoenemann, line 121: Do we really need to move all edges for children? If an edge is contained in the element, its coordinates are usually relative to that element, so we don't need to change them when the element is moved.

jbicker: Yes. They need to be put into `attachedEdgeShifts` because otherwise they aren't handled [here](https://github.com/eclipse-sprotty/sprotty/pull/442/files/6039b06761187ed10c2ca1a6f7dde79f2fccc49a#diff-6c61444caead7afcf404c2d10422199a977d531c2e3cb23625a0cd1af1d2a4d4R187) and the routing points stay where they are.

spoenemann: But what happens if package A contains nodes B and C, and those two nodes are connected by an edge which is also a child of A? Could you try that case? What we need to avoid is moving routing points for elements that are relative to a moved container.

PR #443, opened 2024-03-07
improved check for attached edges

jbicker:
this is a followup for https://github.com/eclipse-sprotty/sprotty/pull/442

### DIFF
--- a/packages/sprotty/src/features/move/move.ts
+++ b/packages/sprotty/src/features/move/move.ts
@@ -112,7 +112,7 @@ export class MoveCommand extends MergeableCommand implements IStoppableCommand {
                     if (this.edgeRouterRegistry) {
                         const handleEdges = (el: SModelElementImpl) => {
                             index.getAttachedElements(el).forEach(edge => {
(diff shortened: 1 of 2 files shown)


packages/sprotty/src/features/move/move.ts
```diff
@@ -211,25 +211,22 @@ export class MoveCommand extends MergeableCommand implements IStoppableCommand {
...
+                    return true;
+                }
+                if (parent instanceof SChildElementImpl) {
+                    return recursiveCheck(parent);
+                }
+                return false;
+            };
+            const isChildOfMovedElement = recursiveCheck(sourceOrTarget);
+
+            return isChildOfMovedElement || Boolean(this.resolvedMoves.get(sourceOrTarget.id));
```
spoenemann, line 226: I'd make use of short-circuit evaluation here:
```suggestion
            return Boolean(this.resolvedMoves.get(sourceOrTarget.id)) || recursiveCheck(sourceOrTarget);
```
if the direct map lookup is positive, we don't need to do the recursive check.

spoenemann (approved):
👍

PR #450, opened 2024-04-29
Fix Junction points post processor

### LINKED ISSUE #451: Issues with junction points
When enabling junction points (using container.load(edgeJunctionModule);) the generated circles are not part of the exported SVG and kept in the visible graph when calling the next time setModel.

2024-04-29_15h15_58
2024-04-29_15h16_33
2024-04-29_15h25_07

gfontorbe:
Fixes #451 

The Junction point postprocessor was too greedy in selecting junction points groups from the DOM. It would also include groups from the hidden model, creating an issue with junction points disappearing when exporting the diagram to svg.

Now the postprocessor is also triggered on the hidden model, and the selectors are adapted to the hidden or base div.

During the first render (when a new model is set) the moved junction points are removed from the DOM.


jbicker (approved):
LGTM

glatuske:
@gfontorbe I have tested your changes locally. Junction points are part of the exported SVG and do not remain in diagram.

PR #454, opened 2024-05-13
SVG export postprocessor

### LINKED ISSUE #421: Self-contained SVG Export
The usage of Sprotty in our project requires to create static SVG screenshots from the graphs rendered with Sprotty. As an additional requirement this needs to be done headless (with node.js only?) and fully self-contained SVGs (on CSS referenced or inlined - all styles need to be part of the corresponding SVG elements).

gfontorbe:
Closes #421 

Adds the possibility to inject a SVG export postprocessor to manipulate the produced SVG element during an export.

Adds the possibility to skip the copying of styles on the exported SVG elements to improve performance.

Fix a bug with the exported viewport calculated on the wrong element.

### DIFF
--- a/packages/sprotty/src/features/export/export.ts
+++ b/packages/sprotty/src/features/export/export.ts
@@ -39,18 +39,22 @@ export class ExportSvgKeyListener extends KeyListener {
     }
 }
 
+export interface ExportSvgOptions {
+    skipCopyStyles?: boolean
+}
+
 export interface RequestExportSvgAction extends RequestAction<ExportSvgAction> {
     kind: typeof RequestExportSvgAction.KIND
-    skipCopyStyles?: boolean
+    options?: ExportSvgOptions
 }
 export namespace RequestExportSvgAction {
     export const KIND = 'requestExportSvg';
 
-    export function create(skipCopyStyles?: boolean): RequestExportSvgAction {
+    export function create(options: ExportSvgOptions = {}): RequestExportSvgAction {
         return {
             kind: KIND,
             requestId: generateRequestId(),
-            skipCopyStyles
+            options
         };
     }
 }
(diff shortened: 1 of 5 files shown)


packages/sprotty/src/features/export/export.ts
```diff
@@ -41,14 +41,16 @@ export class ExportSvgKeyListener extends KeyListener {
 
 export interface RequestExportSvgAction extends RequestAction<ExportSvgAction> {
     kind: typeof RequestExportSvgAction.KIND
+    skipCopyStyles?: boolean
 }
 export namespace RequestExportSvgAction {
     export const KIND = 'requestExportSvg';
 
-    export function create(): RequestExportSvgAction {
+    export function create(skipCopyStyles?: boolean): RequestExportSvgAction {
```
spoenemann, line 49: I would wrap this in an options object because
- passing booleans to a function is not the best API choice in general (`create(true)` can't be understood without looking into the function definition),
- when adding more options to an object, the order doesn't matter and you don't have to pass `undefined` for those that you don't need.

packages/sprotty/src/features/export/svg-exporter.ts
```diff
@@ -14,29 +14,33 @@
...
-    responseId: string
+    svg: string;
+    responseId: string;
+    skipCopyStyles?: boolean;
 }
 export namespace ExportSvgAction {
     export const KIND = 'exportSvg';
 
-    export function create(svg: string, requestId: string): ExportSvgAction {
+    export function create(svg: string, requestId: string, skipCopyStyles: boolean = false): ExportSvgAction {
```
spoenemann, line 38: Same here: wrap the new property in an options object.

spoenemann (commented):
Looks good, just two details

spoenemann (approved):
👍

PR #455, opened 2024-05-17
Fix edge label placement

gfontorbe:
There was a bug where edge labels that do not contain an `edgePlacement` property would be positioned incorrectly


jbicker (approved):
LGTM

PR #458, opened 2024-06-19
Use update boolean instead of hard set 'true' in modelsource submitModel

jbicker:
Before setModel always ended up in UpdateModelAction which triggers (in that case) unwanted animations.


spoenemann (approved):
Remark: the constant `true` value was originally used to keep the previous behavior in [#103](https://github.com/eclipse-sprotty/sprotty/pull/103/files#diff-059b83de3ede5efe2b318c18a8bb3ca39485e32e2128085d696ffba279d0e9f4). The unexpected behavior should be fixed. If you want the model change to be animated, use `updateModel`.

spoenemann:
For reference: the reason for the constant `true` value was that before the introduction of the request/response scheme, there was no easy way to retain the value of that flag during the bounds computation phase:

https://github.com/eclipse-sprotty/sprotty/blob/de20406bd71b62eb1cc34d7ac976befd8019fc33/src/model-source/local-model-source.ts#L238-L255

PR #473, opened 2024-11-20
Adjust binding for `CommandActionHandlerInitializer`

tortmayr:
Make is easier to customize/rebind the
`CommmandActionHandlerInitializer` by binding to itself before binding it as IActionHandlerInitializer.

(All other multiinjected keys already follow this pattern. See https://github.com/eclipse-sprotty/sprotty/issues/239)


spoenemann (approved):
Thanks!

PR #475, opened 2024-12-11
Initial touch support

### LINKED ISSUE #462: Mobile / touch support
Currently, to my knowledge, there is no mobile / touch event support (well mouseDown works, but that's not what I mean).
I think that in particular for viewport manipulation (zoom and scroll), this could be a useful feature. There might be other features which could benefit from this, too (e.g. move).
Some open questions I had when I thought a bit about the feature
- should a one-finger or a two-finger gesture be used for the scroll/pan feature
- when zooming using a pinch gesture, is it allowed to simultaneously scroll the canvas

Maybe it would be an option to - as a first step - add support for touch events in general (either using MouseTool, or using a dedicated TouchTool). For the time being, I know (and am very thankful for) that sprotty is flexible enough that I can implement this myself for my use case, I just thought that maybe others would benefit from such a feature too.

nk-coding:
Adds initial touch support
- `TouchTool` similar to `MouseTool`
- Touch support for canvas scroll / zoom

Open questions
- I did not yet add the `IMouseListener` as discussed in #462 
  - If you want I can do so, however this would likely result in a breaking change, as I would need to change the signature of `MouseTool`
- I added no `decorate` method to `ITouchListener`, as if the `TouchTool` would call it an a class both extends `MouseListener` and `ITouchListener`, it would be called twice
  - I could add it and only call it if the `ITouchListener` is not a `MouseListener`, however this sounds hacky
- I slightly changed the signature of some of the methods of `ScrollMouseListener` (e.g. `moveScrollBar`) to also accept a `Touch` as the event (as this provides the touch coordinates similar to the `MouseEvent`
  - technically, this is a breaking change
  - alternative options: implement touch scroll in separete class (and duplicate some code)
  - add new internal methods to `ScrollMouseListener` which basically are the current ones with the new signature, and make the current one call these new ones
- Currently, there is no touch support for moving elements
  - thus, currently the canvas scroll works even when touching a movable element
  - I can add support for moving elements, however I'm not sure if this is even desired
  - the point above (minor breaking changes vs duplicate code or internal wrapper method) applies here too
 

Closes #462

### DIFF
--- a/packages/sprotty/src/base/di.config.ts
+++ b/packages/sprotty/src/base/di.config.ts
@@ -38,6 +38,7 @@ import { CssClassPostprocessor } from "./views/css-class-postprocessor";
 import { SetModelCommand } from "./features/set-model";
 import { UIExtensionRegistry, SetUIExtensionVisibilityCommand } from "./ui-extensions/ui-extension-registry";
 import { DefaultDiagramLocker } from "./actions/diagram-locker";
+import { TouchTool } from "./views/touch-tool";
 
 const defaultContainerModule = new ContainerModule((bind, _unbind, isBound) => {
     // Logging ---------------------------------------------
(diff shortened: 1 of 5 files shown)


nk-coding:
> One thing I noticed on my MacBook is that the pinch gesture for zooming feels a bit slow.

Sadly, I don't get touch events for that. This results in regular wheel events, which are handled by the `ZoomMouseListener`.
To trigger the touch features, you need to use e.g. a mobile device, or a Laptop with a touch screen.

spoenemann (approved):
Great!

PR #477, opened 2024-12-15
Update inversify

tortmayr:
The inversify dependency in sprotty is locked to 6.0.x. This version is now pretty outdated and there have been a lot of improvements in inversify.js that we can not consume yet.
In addition, this hardlock might block downstream projects.
In GLSP for instance, we might integrate into other components that also use inversify.js (e.g. Theia). 
Unfortunately, dependency injection breaks if multiple different inversify.js versions are resolved so we have to align this.

This change updates to the latest inversify.js version (using 6.1.3 as minimum version same as in Theia)

### DIFF
--- a/examples/package.json
+++ b/examples/package.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "@vscode/codicons": "^0.0.33",
     "express": "^4.18.2",
-    "inversify": "~6.0.2",
+    "inversify": "^6.1.3",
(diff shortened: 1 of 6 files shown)


tortmayr:
@spoenemann 
This inversify version issue currently block a new GLSP release. To fix this we would also need a new (minor) sprotty release.
From your point of view, is there anything preventing a new release after this is merged?

spoenemann (approved):

Issue #485, opened 2025-03-25
Pointer capture support


nk-coding:
Currently, when dragging the viewport (or an element), when the cursor leaves the sprotty area, the dragging process stops until the cursor enters the sprotty area again. This behavior can be changed by using [pointer capture](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture). In my opinion this improves user experience, in particular for split view UIs (e.g. with a langium-based textual editor). In the diagramming context, pointer capture is, for example, used by [React flow](https://reactflow.dev/examples/overview), and [bpmn.io](https://demo.bpmn.io/new).

In case you are interested in such a feature, I would be willing to provide a PR for this. For sprotty, the feature could be built like this:
1. A PointerTool, working similar to the Mouse/TouchTool. Reason: pointer capture is afaik only available with pointer events (the pointer is still captured for mouse events, however for initially setting the pointer capture you need the pointer id)
2. Using a PointerListener instead a MouseListener for the scroll/move handlers. This could be implemented without a breaking change (by still using mouse events for the current logic, and only using the pointerdown event for the capture), however for clean code reasons I would recommend just using the PointerListener and doing this on the next major release.

(in case you want to try this feature with a sprotty-based editor: I implemented it already for https://hylimo.github.io/)

spoenemann:
Yes, sounds good! 👍

PR #488, opened 2025-04-18
Pointer event support

nk-coding:
First half of the implementation for #485 

- adds a `PointerTool`, similar to the already existing `MouseTool` and `TouchTool`
- fixes two small issues from by previous PR #475 
  - renames two parameters for consistency (copy-paste error)
  - exports the `TouchTool` via `index.ts` (I forgot last time, sry)
 
Regarding the second half of #485: this will require replacing sever `MouseListener`s with `PointerListener`s to do pointer capture, which is a breaking change.
I would recommend doing this with the next major release.
I can provide a PR now, or at any in the future based on your notice.

### DIFF
--- a/packages/sprotty/src/base/di.config.ts
+++ b/packages/sprotty/src/base/di.config.ts
@@ -39,6 +39,7 @@ import { SetModelCommand } from "./features/set-model";
 import { UIExtensionRegistry, SetUIExtensionVisibilityCommand } from "./ui-extensions/ui-extension-registry";
 import { DefaultDiagramLocker } from "./actions/diagram-locker";
 import { TouchTool } from "./views/touch-tool";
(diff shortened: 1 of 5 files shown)


spoenemann (approved):
This looks great, thanks!

spoenemann:
> Regarding the second half of https://github.com/eclipse-sprotty/sprotty/issues/485: this will require replacing sever MouseListeners with PointerListeners to do pointer capture, which is a breaking change.
>I would recommend doing this with the next major release.
> I can provide a PR now, or at any in the future based on your notice.

We should start gathering tasks and issues for the [v2.0.0](https://github.com/eclipse-sprotty/sprotty/milestone/12) milestone. We need to ensure that we don't miss other changes that should be done for the major release.

spoenemann:
👍 sounds good!

PR #497, opened 2025-09-26
Added devcontainer.json

spoenemann:
Ona (formerly Gitpod) now uses the devcontainer format. This one was initialized by Codespaces.


gfontorbe (approved):
Tested with Ona and all seems fine

PR #515, opened 2026-03-06
Migration to ESM

gfontorbe:
Migrates the packages to ESM for Sprotty 2.0.0

Upgrading dependencies will come in a future PR

### DIFF
--- a/packages/sprotty/src/base/actions/action-dispatcher.spec.ts
+++ b/packages/sprotty/src/base/actions/action-dispatcher.spec.ts
@@ -14,19 +14,19 @@
  * SPDX-License-Identifier: EPL-2.0 OR GPL-2.0 WITH Classpath-exception-2.0
  ********************************************************************************/
 
-import 'reflect-metadata';
-import { expect, describe, it } from 'vitest';;
 import { Container, injectable, interfaces } from "inversify";
-import { Action, RedoAction, RejectAction, RequestModelAction, SetModelAction, UndoAction } from 'sprotty-protocol/lib/actions';
-import { Bounds } from 'sprotty-protocol/lib/utils/geometry';
-import { TYPES } from "../types";
-import { InitializeCanvasBoundsAction } from '../features/initialize-canvas';
-import { Command, CommandExecutionContext, CommandReturn, ICommand } from '../commands/command';
-import { ICommandStack } from "../commands/command-stack";
-import { ActionDispatcher } from "./action-dispatcher";
-import defaultModule from "../di.config";
-import { EMPTY_ROOT } from '../model/smodel-factory';
-import { IActionHandler, configureActionHandler } from './action-handler';
+import 'reflect-metadata';
+import { Action, Bounds, RedoAction, RejectAction, RequestModelAction, SetModelAction, UndoAction } from 'sprotty-protocol';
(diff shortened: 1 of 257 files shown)


packages/sprotty/src/base/actions/action-dispatcher.spec.ts
```diff
@@ -14,19 +14,19 @@
...
+import { describe, expect, it } from 'vitest';
+import { ICommandStack } from "../commands/command-stack.js";
+import { Command, CommandExecutionContext, CommandReturn, ICommand } from '../commands/command.js';
+import defaultModule from "../di.config.js";
+import { InitializeCanvasBoundsAction } from '../features/initialize-canvas.js';
+import { EMPTY_ROOT } from '../model/smodel-factory.js';
+import { TYPES } from "../types.js";
+import { ActionDispatcher } from "./action-dispatcher.js";
+import { IActionHandler, configureActionHandler } from './action-handler.js';
+;
```
kaisalmen: That is left over

packages/sprotty/src/base/views/view.spec.ts
```diff
@@ -14,17 +14,16 @@
...
 import { Container } from 'inversify';
-import { TYPES } from '../types';
-import { EMPTY_ROOT, SModelFactory } from '../model/smodel-factory';
-import { SNodeImpl } from "../../graph/sgraph";
-import { EmptyView, findArgValue, MissingView } from "./view";
-import { ModelRenderer } from "./viewer";
-import defaultModule from "../di.config";
-
-const toHTML = require('snabbdom-to-html');
+import 'reflect-metadata';
```
kaisalmen: Wasn't it required to import reflect-metadata always first? There is something in the back of my head that makes me think that. If, true, 
that is applicable for all those imports in all other places.

kaisalmen (commented):
Looking good already. I have some suggestions. I did not check each and every import change and trust that VS Code's import clean-up and re-order does its job corrrectly.

We could have a direct chat tomorrow. 🙂

packages/sprotty/src/index.ts
```diff
@@ -16,169 +16,169 @@
...
-export * from './base/views/viewer-cache';
-export * from './base/views/viewer-options';
-export * from './base/views/viewer';
-export * from './base/views/vnode-postprocessor';
-export * from './base/views/vnode-utils';
-
-export * from './base/types';
-
-import defaultModule from './base/di.config';
+export * from './base/actions/action.js';
```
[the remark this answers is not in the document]
spoenemann: Many of the subfolders are very small and the upside of having more topic-related imports is also a downside: you have to know from which sub-export to get the thing you want (unless auto-import works properly). We wouldn't save anything in terms of bundle size, as the majority of the code base is referenced anyway (e.g. through [modules.ts](https://github.com/eclipse-sprotty/sprotty/blob/master/packages/sprotty/src/lib/modules.ts)).

But one thing we could consider is adding a sub-export for `features`. That would help to indicate which things are not strictly necessary.

spoenemann (commented):
Thanks a lot for this effort @gfontorbe!

Some of the examples are currently broken: Micro-Layout Showcase, Class Diagram, Random Graph, SVG

spoenemann (approved):
🎉

PR #534, opened 2026-06-05
Migration to npm

gfontorbe:
Removes yarn in favor of npm and adapts scripts and configs where appropriate


.github/workflows/build.yml
```diff
@@ -22,13 +22,15 @@ jobs:
       uses: actions/setup-node@v4
       with:
         node-version: '24'
+    - name: Update npm
+      run: npm install -g npm@latest
```
spoenemann: Is this necessary? If yes, shall we pin the version number?

package.json
```diff
@@ -1,11 +1,11 @@
 {
   "private": true,
   "scripts": {
-    "prepare": "yarn run clean && yarn run build",
+    "prepare": "npm run clean && npm run build",
```
spoenemann: Suggestion: remove the prepare script. In other projects, `install` really only installs dependencies. Building should be a separate step.

.devcontainer/devcontainer.json
```diff
@@ -12,7 +12,7 @@
 	"forwardPorts": [8080],
 
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "yarn install"
+	"postCreateCommand": "npm install"
```
spoenemann: After removing the prepare script, this should become `npm install && npm run build`

package.json
```diff
@@ -1,11 +1,11 @@
...
     "clean": "lerna run clean",
```
spoenemann: This can be replaced with `npm run clean --workspaces`. I think lerna is only used for publishing – we can kick it out later.

spoenemann (commented):
Thanks!

spoenemann (approved):

PR #540, opened 2026-07-13
Fix exported SVG block sizing

oiahoon:
## Summary

- exclude the computed `block-size` when copying root styles into exported SVGs
- keep unrelated computed styles and child style copying unchanged
- add dedicated regression coverage for every excluded root property

## Testing

- focused SVG exporter regression: 5 passed
- TypeScript project build: passed
- ESLint on changed files: passed
- full Vitest suite: 210/223 passed; untouched `master` is 205/218 with the same 13 JSX assertion failures caused by the current transform injecting `__self` and `__source` attributes

Fixes #520

### DIFF
--- a/packages/sprotty/src/features/export/svg-exporter.spec.ts
+++ b/packages/sprotty/src/features/export/svg-exporter.spec.ts
@@ -1,0 +1,84 @@
+/********************************************************************************
+ * Copyright (c) 2026 TypeFox and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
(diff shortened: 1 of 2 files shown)


gfontorbe:
Hi @oiahoon, thanks for the contribution.

Would you mind adding tests for the other excluded properties as well? Each should get it's own dedicated test case.

Also, please sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/eca/) via your Eclipse Foundation account in order for this PR to be considered.

oiahoon:
> Hi @oiahoon, thanks for the contribution.
> 
> Would you mind adding tests for the other excluded properties as well? Each should get it's own dedicated test case.
> 
> Also, please sign the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/eca/) via your Eclipse Foundation account in order for this PR to be considered.

surely.

oiahoon:
Addressed in 39732ad. Added dedicated tests for width, height, opacity, inline-size, and block-size; the focused tests, TypeScript build, and ESLint pass. The ECA check is also passing now.

gfontorbe:
@oiahoon Thank you very much

PR #542, opened 2026-07-15
Fix stale bounds leaking into ComputedBoundsAction after SVG export

### LINKED ISSUE #541: HiddenBoundsUpdater leaks stale bounds into the next ComputedBoundsAction after an SVG export
HiddenBoundsUpdater.decorate() collects bounds data during every hidden rendering, but postUpdate() only cleans the collected map up when the rendering was caused by a RequestBoundsAction — for any other cause it returns early without clearing:

https://github.com/eclipse-sprotty/sprotty/blob/master/packages/sprotty/src/features/bounds/hidden-bounds-updater.ts#L75-L78

An SVG export (ExportSvgCommand is a HiddenCommand, cause RequestExportSvgAction) is exactly such a rendering: it fills the map with entries for the cloned export root, which are never removed. The next RequestBoundsAction then produces a ComputedBoundsAction that additionally contains entries for every element of the stale export clone — including elements that no longer exist in the model.

Steps to reproduce:

1. Trigger an SVG export
2. Remove an element from the model (or change element ids)
3. The next bounds computation sends a ComputedBoundsAction referencing the removed element

Impact: LocalModelSource masks the problem because applyBounds silently skips unknown ids, lementations receive the bogus action and may reject the whole model update. Observeddownstream in GLSP, where the server responds with GLSPServerException: Model element not f1717). Even without deletions, every ComputedBoundsAction after an export carries duplicateentries for all surviving elements.

Fix proposal: reset the collected data in postUpdate regardless of the rendering cause

tortmayr:
Reset the collected bounds data after every hidden rendering, so renderings that
don't compute bounds (e.g. SVG exports) no longer leak stale entries into the
next `ComputedBoundsAction`.

Fixes #541

### DIFF
--- a/packages/sprotty/src/features/bounds/hidden-bounds-updater.spec.ts
+++ b/packages/sprotty/src/features/bounds/hidden-bounds-updater.spec.ts
@@ -1,0 +1,104 @@
+/********************************************************************************
+ * Copyright (c) 2026 EclipseSource and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
+ * Licenses when the conditions for such availability set forth in the Eclipse
+ * Public License v. 2.0 are satisfied: GNU General Public License, version 2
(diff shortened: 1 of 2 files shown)


spoenemann (approved):
Thanks!

PR #548, opened 2026-08-20
replace `innerHtml` incidences

Wagner-Laranjeiras:
Command palette suggestions built labels and icons by concatenating HTML strings. Action labels can come from diagram/model data, so that was a security risk.

This change builds the same DOM with createElement, textContent, and append. Search highlighting still wraps matches in <em>, and spaces are preserved as non-breaking spaces.


spoenemann (approved):
Thanks!

PR #566, opened 2026-09-03
Remove deprecated code

gfontorbe:
Remove deprecated code from the codebase. Cleanup before v2.0

### DIFF
--- a/packages/sprotty/src/features/viewport/viewport.ts
+++ b/packages/sprotty/src/features/viewport/viewport.ts
@@ -104,15 +104,11 @@ export class GetViewportCommand extends ModelRequestCommand {
 
 export class ViewportAnimation extends Animation {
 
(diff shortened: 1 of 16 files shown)


packages/sprotty/src/features/viewport/viewport.ts
```diff
@@ -104,15 +104,11 @@ export class GetViewportCommand extends ModelRequestCommand {
 
 export class ViewportAnimation extends Animation {
 
-    /** @deprecated Use `Math.log(newViewport.zoom / oldViewport.zoom)` instead. */
-    protected zoomFactor: number;
```
spoenemann, line 108: 🤔 should we provide a getter property for this?

spoenemann (approved):
Thanks!

